PR #3, opened 2022-06-02
SymPy unit tests

hsiang-wu:
Unit tests of SymPy showing its basic usages.

New file `simple_test.py` contains simple tests of SymPy functions.
New file `simplification_test` showcases simplification-related functions of SymPy.

### DIFF
--- a/neuralpp/test/quick_tests/symoblic/simple_test.py
+++ b/neuralpp/test/quick_tests/symoblic/simple_test.py
@@ -2,7 +2,10 @@
 Tests of features covered in the following tutorials of SymPy.
 https://docs.sympy.org/latest/tutorial/intro.html
 https://docs.sympy.org/latest/tutorial/gotchas.html
+https://docs.sympy.org/latest/tutorial/basic_operations.html
 """
+import math
+
 from sympy import *
 
 
@@ -54,18 +57,37 @@ def test_change_after_create() -> None:
     assert expr == symbols('x') + 1
 
 
-def test_xor() -> None:
-    """ ^ in SymPy is reserved for xor (not exponentiation), as is in Python."""
+def test_gotchas_xor_and_divide() -> None:
+    """Some gotchas of the library: ^ is xor, / can be float division. """
+    # ^ in SymPy is reserved for xor (not exponentiation), as is in Python.
     x, y = symbols('x y')
     assert x ^ y == Xor(x, y)
     assert x ^ y != x ** y
 
-
-def test_divide() -> None:
-    """ / in Python3 is float division. So SymPy uses Rational() explicitly."""
+    # / in Python3 is float division. So SymPy uses Rational() explicitly.
     # the following tests would pass as the rational 1/2 equals the division result 1/2=0.5
     assert Rational(1, 2) == 1/2
     assert Rational(1, 2) == Integer(1)/Integer(2)
     # things are a little different when it comes to 1/3
     assert Rational(1, 3) != 1/3
     assert Rational(1, 3) == Integer(1)/Integer(3)
+
+
+def test_substitution():
+    """ Test variable substitution. """
+    x, y, z = symbols("x y z")
+    expr = cos(x) + 1
+    assert expr.subs(x, y) == cos(y) + 1
+    assert expr.subs(x, x**y) == cos(x**y) + 1
+
+
+def test_eval():
+    """ Test numerical evaluation, including a quantifier example. """
+    x, i = symbols("x i")
+    assert sqrt(8) != math.sqrt(8)
+    assert sqrt(8).evalf() == math.sqrt(8)
+    assert cos(2*x).evalf(subs={x: 2.4}) == math.cos(2*2.4)
+
+    expr = Sum(Indexed('x', i), (i, 0, 3))  # expr is a quantified expression
+    expr_as_func = lambdify(x, expr)
+    assert expr_as_func([1, 2, 3, 4, 5]) == 10
--- a/neuralpp/test/quick_tests/symoblic/simple_test.py
+++ b/neuralpp/test/quick_tests/symoblic/simple_test.py
@@ -9,7 +9,7 @@ import math
 # cannot use `from sympy import *` because that would import all the "test_*" functions in sympy,
 # causing pytest to run them as well.
 from sympy import symbols, expand, factor, sin, cos, diff, integrate, exp, Integer, Xor, Rational, \
-    sqrt, Sum, lambdify, Indexed, factorial, oo
+    sqrt, Sum, lambdify, Indexed, factorial, oo, Integral
 
 
 def test_symbols_basic():
@@ -118,11 +118,13 @@ def test_sum():
 
     # `i` can also be used in the series. An alternative way to the above example is the following.
     expr = Sum(i, (i, 0, 100))  # expr is a quantified expression
-    # Here doit() does evaluation (e.g., it "does" the sum), note the interval can be infinite
+    # Here doit() does symbolic evaluation (e.g., it "does" the sum), note the interval can be infinite
     assert expr.doit() == 5050
 
     # Note doit() does *symbolic* evaluation. In the above example we get an integer only because there's no symbol.
     assert Sum(x**k/factorial(k), (k, 0, oo)).doit() == exp(x)
+    # You can also call doit() on integral. It makes sense because we can "do" integral.
+    assert (2 * Integral(x, x)).doit() == x**2
     # Doc in Sum() defines the following behavior which is a little weird.
     # If start > end in (i, start, end), it is defined to be the same as (i, end+1, start-1)
     # https://docs.sympy.org/latest/modules/concrete.html?highlight=sum#sympy.concrete.summations.Sum
--- a/neuralpp/test/quick_tests/symoblic/simple_test.py
+++ b/neuralpp/test/quick_tests/symoblic/simple_test.py
@@ -6,7 +6,7 @@ https://docs.sympy.org/latest/tutorial/basic_operations.html
 """
 import math
 
-# cannot use from sympy import * because that would import all the "test_*" functions in sympy,
+# cannot use `from sympy import *` because that would import all the "test_*" functions in sympy,
 # causing pytest to run them as well.
 from sympy import symbols, expand, factor, sin, cos, diff, integrate, exp, Integer, Xor, Rational, \
     sqrt, Sum, lambdify, Indexed, factorial, oo
--- a/neuralpp/test/quick_tests/symoblic/simple_test.py
+++ b/neuralpp/test/quick_tests/symoblic/simple_test.py
@@ -12,7 +12,7 @@ from sympy import symbols, expand, factor, sin, cos, diff, integrate, exp, Integ
     sqrt, Sum, lambdify, Indexed, factorial, oo
 
 
-def test_symbols_basic() -> None:
+def test_symbols_basic():
     """SymPy supports simple expression conversions. e.g., `x + 2 * y` is the same as `y + x + y`"""
     x, y = symbols('x y')
     expr = x + 2 * y
@@ -20,7 +20,7 @@ def test_symbols_basic() -> None:
     assert expr == y + x + y
 
 
-def test_expand_and_factor() -> None:
+def test_expand_and_factor():
     """
     The simplification SymPy automatically performs is very limited. For example,
     x(x+2y) is not "equal" in the sense of SymPy's symbolic representation as x^2+2xy.
@@ -35,7 +35,7 @@ def test_expand_and_factor() -> None:
     assert factor_form_expr == factor(expand_form_expr)
 
 
-def test_computation() -> None:
+def test_computation():
     """ Symbolic computation """
     x, t, z, nu = symbols('x t z nu')
     # for simple formulas integrate and diff are invertible
@@ -52,7 +52,7 @@ def test_computation() -> None:
     assert complex_formula != x + (x-1) * exp(x) + exp(x)
 
 
-def test_change_after_create() -> None:
+def test_change_after_create():
     """changing the binding of python variable does not affect expression already created """
     x = symbols('x')
     expr = x + 1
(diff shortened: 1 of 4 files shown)


neuralpp/test/quick_tests/symoblic/symbols_test.py
```diff
@@ -0,0 +1,71 @@
...
+https://docs.sympy.org/latest/tutorial/gotchas.html
+"""
+from sympy import *
+
+
+def test_symbols_basic() -> None:
+    """SymPy supports simple expression conversions. e.g., `x + 2 * y` is the same as `y + x + y`"""
+    x, y = symbols('x y')
+    expr = x + 2 * y
+    assert expr == x + 2 * y
```
rodrigodesalvobraz: Interesting! Does that mean the operator == does normalization work every time, or are expressions immediately converted to a normal form as soon as they are created so that equivalent functions are represented identically internally? Does the manual say anything about that? I guess one way to find out is to create two equivalent but distinct expressions and print them to see if their string representation is the same (which would mean they are normalized and the initial form is lost).

neuralpp/test/quick_tests/symoblic/simplication_test.py
```diff
@@ -0,0 +1,54 @@
...
+Tests of SymPy's simplication functions.
+https://docs.sympy.org/latest/tutorial/simplification.html
+"""
+from sympy import *
+
+
+def test_simplify() -> None:
+    """
+    One can just use sympy.simplify().
+    It's push-button but has 2 caveats:
```
rodrigodesalvobraz: What does "push-down" mean?

neuralpp/test/quick_tests/symoblic/simplication_test.py
```diff
@@ -0,0 +1,54 @@
...
+    1. "simplest" form is not well-defined. (showcased by last set of asserts)
+    2. it can be unnecessarily slow due to generality.
+    """
+    x = symbols('x')
+    assert simplify(sin(x)**2 + cos(x)**2) == 1
+    assert simplify((x**3 + x**2 - x - 1)/(x**2 + 2*x + 1)) == x - 1
+    assert simplify(gamma(x) / gamma(x - 2)) == (x - 2) * (x - 1)
+    # we may want to simplify (x**2 + 2*x + 1) into (x+1)**2 but that's not "simplest" recognized by simplify()
```
rodrigodesalvobraz, line 19: It's very nice that you are using the tests to illustrate the finer points like that!

neuralpp/test/quick_tests/symoblic/simplication_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+def test_expand() -> None:
+    """ The function expand() expands. """
+    x = symbols('x')
+    assert expand((x + 2)**2) == x**2 + 4*x + 4
+    assert expand((x + 2)*(x - 2)) == x**2 - 4
+
+
+def test_factor() -> None:
+    """ The function factor() turns a polynomial into an irreducible product of factors. """
```
rodrigodesalvobraz: Does it work for arbitrary degrees? I don't know much about this but my understanding is that this is hard for higher degrees.

neuralpp/test/quick_tests/symoblic/simple_test.py
```diff
@@ -54,18 +57,37 @@ def test_change_after_create() -> None:
...
+
+
+def test_eval():
+    """ Test numerical evaluation, including a quantifier example. """
+    x, i = symbols("x i")
+    assert sqrt(8) != math.sqrt(8)
+    assert sqrt(8).evalf() == math.sqrt(8)
+    assert cos(2*x).evalf(subs={x: 2.4}) == math.cos(2*2.4)
+
+    expr = Sum(Indexed('x', i), (i, 0, 3))  # expr is a quantified expression
```
rodrigodesalvobraz, line 94: Nice that you tested a quantifier! But I don't quite get the meaning here: is i the index? How is x being used exactly? What is the meaning of (i, 0, 3)?

neuralpp/test/quick_tests/symoblic/simplication_test.py
```diff
@@ -22,33 +21,53 @@ def test_simplify() -> None:
     assert factor(x**2 + 2*x + 1) == (x + 1)**2
 
 
-def test_expand() -> None:
-    """ The function expand() expands. """
-    x = symbols('x')
+def test_polynomial_rational() -> None:
```
rodrigodesalvobraz, line 25: The original name test_expand seemed clearer to me, I guess because "expand" is an action and we test actions. What does it mean to test a polynomial_rational?

neuralpp/test/quick_tests/symoblic/simple_test.py
```diff
@@ -83,11 +83,16 @@ def test_substitution():
...
     assert sqrt(8) != math.sqrt(8)
     assert sqrt(8).evalf() == math.sqrt(8)
     assert cos(2*x).evalf(subs={x: 2.4}) == math.cos(2*2.4)
 
     expr = Sum(Indexed('x', i), (i, 0, 3))  # expr is a quantified expression
-    expr_as_func = lambdify(x, expr)
+    expr_as_func = lambdify(x, expr)  # lambdify() turns the expression into a python function
     assert expr_as_func([1, 2, 3, 4, 5]) == 10
+
+    # doit() can also do computation/simplification, note the interval can be infinite
```
rodrigodesalvobraz, line 98: What is the different between doit and simplification? When should we use one or the other?

neuralpp/test/quick_tests/symoblic/simple_test.py
```diff
@@ -73,15 +76,15 @@ def test_gotchas_xor_and_divide() -> None:
     assert Rational(1, 3) == Integer(1)/Integer(3)
 
 
-def test_substitution():
+def test_substitution() -> None:
```
rodrigodesalvobraz, line 79: Why did you add "-> None"? Was that required?

neuralpp/test/quick_tests/symoblic/symbols_test.py
hsiang-wu: > Does that mean the operator == does normalization work every time, or are expressions immediately converted to a normal form as soon as they are created so that equivalent functions are represented identically internally?

I remember I read from the doc somewhere that it is the latter, where they gave the example of `x + 1` and `1 + x` are printed in the same way. And for `==`, it is said that "in SymPy, == represents exact structural equality testing."

neuralpp/test/quick_tests/symoblic/simplication_test.py
hsiang-wu: Neither do I. From the tutorial:

> factor() uses a complete multivariate factorization algorithm over the rational numbers, which means that each of the factors returned by factor() is guaranteed to be irreducible.

So my guess is it works for arbitrary degrees though it might be slower for higher degrees.

neuralpp/test/quick_tests/symoblic/simple_test.py
rodrigodesalvobraz: Yes, I think that deserves some more detail (enough to understand how to use it ourselves).

neuralpp/test/quick_tests/symoblic/simplication_test.py
hsiang-wu: A better name may be `test_polynomial_or_rational_function_simplification`. The library has a dedicated set of simplification functions for each kind of expressions, e.g., `expand()` and `factor()` for polynomial/rational, `trigsimp()` for trigonometric.

neuralpp/test/quick_tests/symoblic/simple_test.py
hsiang-wu: I should have avoid using the word "simplification". A better word would be "evaluation". Since the expression is symbolic, `Sum` is just `Sum`, they don't realy "do" the sum. Calling `doit()` would actually do the sum.

So calling `doit()` makes sense on the expressions that you can "do", such as `Sum` and `Integral`. To use another example from the doc, `2*Integral(x, x)` is just `2*Integral(x, x)`, but `(2*Integral(x, x)).doit()` would do the integral and result in `x**2`.

neuralpp/test/quick_tests/symoblic/simple_test.py
hsiang-wu: No. Just want to write it in a more typed way..

neuralpp/test/quick_tests/symoblic/simple_test.py
```diff
@@ -0,0 +1,101 @@
...
+def test_eval() -> None:
+    """ Test numerical evaluation, including a quantifier example. """
+    x, i, k = symbols("x i k")
+    assert sqrt(8) != math.sqrt(8)
+    assert sqrt(8).evalf() == math.sqrt(8)
+    assert cos(2*x).evalf(subs={x: 2.4}) == math.cos(2*2.4)
+
+    expr = Sum(Indexed('x', i), (i, 0, 3))  # expr is a quantified expression
+    expr_as_func = lambdify(x, expr)  # lambdify() turns the expression into a python function
+    assert expr_as_func([1, 2, 3, 4, 5]) == 10
```
CactusWin, line 96: Is there a reason that you included `5` in the list,  since it won't be reached in the expression? Was it to show that it'll ignore extra arguments?

CactusWin (commented):
Very nice! I really liked the comments you made! It made the test much easier to read and understand. :) I'll approve once the changes that Rodrigo suggested are made/ resolved.

neuralpp/test/quick_tests/symoblic/simple_test.py
rodrigodesalvobraz: Yes, I think it may be distracting since the rest of the code is not typed.

neuralpp/test/quick_tests/symoblic/simplication_test.py
rodrigodesalvobraz: Ah, now I understand. I think many people may get confused like I did, so can you please add a short explanation ("automatic and general", something like that)?

neuralpp/test/quick_tests/symoblic/simplication_test.py
rodrigodesalvobraz: Can you please update the name then? Thanks.

neuralpp/test/quick_tests/symoblic/simple_test.py
rodrigodesalvobraz: Can you please add some of that explanation to the comments? Thanks.

horizon-blue (commented):
:D I really like the detailed comments. The two test files looks great to me overall minus Rodrigo's and Winnie's suggestions. Once we address those, the PR should be good to go.

neuralpp/test/quick_tests/symoblic/simple_test.py
hsiang-wu: Yes. To show `i` indexes from 0 to 3 and not beyond.

neuralpp/test/quick_tests/symoblic/simplication_test.py
hsiang-wu: I replaced "push-button" with "general", does that make sense?

neuralpp/test/quick_tests/symoblic/simple_test.py
hsiang-wu: Created a new test function for `Sum()`

neuralpp/test/quick_tests/symoblic/simplication_test.py
hsiang-wu: Changed the function names in this file.

neuralpp/test/quick_tests/symoblic/simple_test.py
hsiang-wu: Refactored test for `Sum()` and moved test of `doit()` there;
switched to the term "symbolic evaluation" in the comment.

neuralpp/test/quick_tests/symoblic/simple_test.py
hsiang-wu: Got rid of `-> None` in test files.

CactusWin:
This looks good to me! I'm not too sure how to approve the PR, since the review button seemed to have disappeared for me :)

horizon-blue (approved):
@CactusWin I think you can either re-request review for your account from the right side panel, or visit the "Files changed" tab to directly create a new review :)

neuralpp/test/quick_tests/symoblic/simple_test.py
```diff
@@ -0,0 +1,131 @@
...
+def test_sum():
+    """ A more detailed test of the Sum quantifier.
+    In quote of Sum's doc, the first argument is "the general form of terms in the series" and
+    the second argument is "(dummy_variable, start, end), with dummy_variable taking all
+    integer values from start through end. In accordance with long-standing mathematical convention,
+    the end term is included in the summation."
+    """
+    x, i, k = symbols("x i k")
+    # In the following example, `i` is not used in the series but just to indicate the index of 'x'.
+    expr = Sum(Indexed('x', i), (i, 0, 100))
```
rodrigodesalvobraz, line 115: I think this is a bit too confusing for a first example. I thought that somehow `Indexed` referred to the summation's index, not that `x` was a list being indexed. Also, I got confused by your comment that `i` is not used in the series, because it is (to index `x`). It would be good to start with a plainer example like `Sum(2*i, (i, 0, 100))` and then move on to the use of `Indexed`.

rodrigodesalvobraz (changes requested):
Looking good! Please address a small change I suggested for the Sum function and then we should merge. Thanks!

neuralpp/test/quick_tests/symoblic/simple_test.py
hsiang-wu: Got it. I'v reordered tests and modified comments.

rodrigodesalvobraz (approved):

PR #5, opened 2022-06-08
Implemented Expression Interface

hsiang-wu:
In addition to the project plan, I also implemented a `replace()` method, which is similar to a common API of both Z3 and SymPy.

I did not choose to implement `Variable`, `Constant` and `FunctionApplication` as abstract because their attributes and properties can be shared among subclasses. In that case, `BasicVariable/Constant/FunctionApplication` are just empty vessels.

I also included a test `pattern_match_test` for Python's new pattern matching feature which I'm not very familiar with. I found its official doc a little ambiguous so I played with it a bit.

Pattern matching is used in the implementation of `__eq__()`. It can also be useful when implementing interpreter, etc.


neuralpp/symbolic/expression.py
```diff
@@ -0,0 +1,151 @@
...
+    @abstractmethod
+    def replace(self, from_expression: 'Expression', to_expression: 'Expression') -> 'Expression':
+        """
+        Every expression is immutable so replace() returns either self or a new Expression.
+        No in-place modification should be made.
+        If from_expression is not in the expression, returns self.
+        """
+        pass
+
+    def contains(self, target: 'Expression') -> bool:
```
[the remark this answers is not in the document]
hsiang-wu: Yeah maybe that makes more sense. In that case `Expression` can be purely abstract.

neuralpp/symbolic/expression.py
```diff
@@ -0,0 +1,151 @@
+from typing import Any, List
+from abc import ABC, abstractmethod
+
+
+class Expression(ABC):
+    @abstractmethod
+    # 'Expression': forward reference for recursive type reference https://peps.python.org/pep-0484/#forward-references
+    def subexpression(self) -> List['Expression']:
```
horizon-blue, line 8: Looks like `subexpression` could be a property instead of a method, what do you think? ;)

neuralpp/symbolic/expression.py
```diff
@@ -0,0 +1,151 @@
+from typing import Any, List
+from abc import ABC, abstractmethod
+
+
+class Expression(ABC):
+    @abstractmethod
+    # 'Expression': forward reference for recursive type reference https://peps.python.org/pep-0484/#forward-references
```
horizon-blue, line 7: FYI: you can use `from __future__ import annotations` to [postpone the evaluation of annotations](https://stackoverflow.com/questions/61544854/from-future-import-annotations), which should allow you to use `Expression` in type hinting (without using the string)

neuralpp/symbolic/function.py
```diff
@@ -0,0 +1,46 @@
+from typing import Callable
+from abc import ABC, abstractmethod
+
+
+class Function(ABC):
+    """
+    The motivation of having a dedicated Function class is to support `built-in functions`.
```
rodrigodesalvobraz, line 7: I am not quite seeing the argument here, or the need for Function(). If we use `BasicConstant(operator.add)` then we can recognize `operator.add` as a supported function and do the same things we could do with `Add()`, but without having to define `Function` and `Add`, isn't it so?

neuralpp/test/quick_tests/symoblic/expression_test.py
```diff
@@ -0,0 +1,84 @@
...
+
+def test_basic_constant():
+    # Constant can be anything
+    constant_one = BasicConstant(1)
+    constant_abc = BasicConstant("abc")
+    assert constant_one != constant_abc
+    assert constant_one == BasicConstant(2-1)
+    assert constant_one.subexpression() == []
+    assert not constant_one.contains(constant_abc)
+    assert not constant_one.contains(constant_one)  # A constant does not contains() itself
```
rodrigodesalvobraz, line 15: I was thinking of the semantics of `contains` as similar to "subset". A set contains itself, so an expression should contain itself.

neuralpp/test/quick_tests/symoblic/expression_test.py
```diff
@@ -0,0 +1,84 @@
...
+    assert constant_one.replace(constant_one, constant_abc) == constant_one  # Any replace() has no effect on constant
```
rodrigodesalvobraz, line 16: Likewise, I think replace should replace constants too. If I say: "replace every occurrence of 1 by 2 in the following expression: 1", I think the answer should be 2, no?

neuralpp/test/quick_tests/symoblic/expression_test.py
```diff
@@ -0,0 +1,84 @@
...
+def test_basic_variable():
+    # Variable is initialized by a variable name.
+    variable_x = BasicVariable("x")
+    variable_y = BasicVariable("y")
+    assert variable_x != variable_y
+    assert variable_x == BasicVariable("x")
+    assert variable_x.subexpression() == []
+    assert not variable_x.contains(variable_y)
+    assert not variable_x.contains(variable_x)  # A variable does not contains() itself
+    assert variable_x.replace(variable_x, variable_y) == variable_x  # Any replace() has no effect on variable
```
rodrigodesalvobraz, line 30: Again here, I think replacing x by y in x should return y.

neuralpp/test/quick_tests/symoblic/expression_test.py
```diff
@@ -0,0 +1,84 @@
...
+    assert fa3.contains(constant_two)  # shows that search of contains() is deep.
+
+    # replace() is also deep and not in-place (returns a new object instead of modify the called on)
+    fa4 = fa3.replace(constant_two, constant_one)
+    assert fa4 != fa3
+    assert fa3 == BasicFunctionApplication(func3, [constant_one, fa2])
+    assert fa4 == BasicFunctionApplication(func3, [constant_one,
+                                                   BasicFunctionApplication(func2, [constant_one, constant_one])])
+
+    # set() is also not changing the object that's called on
```
rodrigodesalvobraz, line 77: nit: "that it is called on"

neuralpp/test/quick_tests/symoblic/expression_test.py
```diff
@@ -0,0 +1,84 @@
...
+    fa5 = fa2.set(0, BasicVariable("f"))  # Any replace() has no effect on variable
```
rodrigodesalvobraz, line 78: Is the comment about "replace" left over here from a copy and paste? It does not seem to apply.

neuralpp/symbolic/expression.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def replace(self, from_expression: Expression, to_expression: Expression) -> Expression:
+        return self
+
+    def set(self, i: int, new_expression: Expression) -> Expression:
+        raise IndexError("Variable has no subexpression, so you cannot set().")
+
+    def __eq__(self, other):
+        match other:
+            case Variable(name=other_name):
```
horizon-blue, line 70: :D Ah, Python's structural pattern matching is so clean. AFAIK it's only available since Python 3.10, though. @rodrigodesalvobraz do you have any concern about compatibility with older version of Python?

rodrigodesalvobraz (changes requested):
Cool stuff! I think this will require a few more changes but overall it's in a good shape!

neuralpp/symbolic/expression.py
```diff
@@ -0,0 +1,151 @@
...
+                return self._name == other_name
+            case _:
+                return False
+
+
+class Constant(Expression):
```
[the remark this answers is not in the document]
hsiang-wu: I agree.

neuralpp/symbolic/expression.py
```diff
@@ -0,0 +1,151 @@
...
+        If from_expression is not in the expression, returns self.
+        """
+        pass
+
+    def contains(self, target: 'Expression') -> bool:
+        """
+        Checks if `target` is contained in `self`. The check is deep. E.g.,
+        f(x,f(a,b)).contains(a) == True
+        """
+        for sub_expr in self.subexpression():
```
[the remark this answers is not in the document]
hsiang-wu: I see. That's a different `contains()` semantics. I'll change to that.

neuralpp/test/quick_tests/symoblic/expression_test.py
hsiang-wu: got it.

neuralpp/test/quick_tests/symoblic/expression_test.py
hsiang-wu: fixed

neuralpp/symbolic/expression.py
hsiang-wu: Good to know! Thanks!

neuralpp/symbolic/expression.py
hsiang-wu: I think we can do that, since `subexpression` is not recursive and thus not computation-heavy. It's not specified as a call instead of a property in the project plan. @rodrigodesalvobraz do you think we should change that?

neuralpp/symbolic/expression.py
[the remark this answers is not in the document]
hsiang-wu: I think we want to have `BasicConstant("a") != BasicVariable("a")` and `BasicConstant("a") == SymPyConstant("a")`, so we may still want to stick to this pattern. Alternatively, I can define an `base_type` interface in `AtomicExpression` and use that for comparison instead of `type`

neuralpp/symbolic/expression.py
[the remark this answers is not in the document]
hsiang-wu: per @rodrigodesalvobraz's suggestion of a new `contains()` semantics, now it might be better to have `contains()` in `Expression`? Since `contains()` of `AtomicExpression` and `FunctionApplication` share the `if self == target: return True` logic, having a `contains()` in `Expression` is more DRY?

neuralpp/symbolic/expression.py
```diff
@@ -0,0 +1,151 @@
...
+        return self._name
+
+    def subexpression(self) -> List[Expression]:
+        return []
+
+    def replace(self, from_expression: Expression, to_expression: Expression) -> Expression:
+        return self
+
+    def set(self, i: int, new_expression: Expression) -> Expression:
+        raise IndexError("Variable has no subexpression, so you cannot set().")
```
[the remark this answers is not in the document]
hsiang-wu: got this

neuralpp/symbolic/expression.py
```diff
@@ -0,0 +1,151 @@
...
+            case Constant(value=other_value):
+                return self._value == other_value
+            case _:
+                return False
+
+
+class FunctionApplication(Expression):
+    __match_args__ = ("function", "arguments")
+
+    def __init__(self, func: Expression, args: List[Expression]):
```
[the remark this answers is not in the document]
hsiang-wu: got this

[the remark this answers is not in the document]
hsiang-wu: I get the general idea, but for `args` specifically, it is kind of a term? I see some of its usages in neuralpp also.

neuralpp/symbolic/expression.py
```diff
@@ -0,0 +1,151 @@
...
+
+    @property
+    def function(self) -> Expression:
+        return self._subexpression[0]
+
+    @property
+    def arguments(self) -> List[Expression]:
+        return self._subexpression[1:]
+
+    def subexpression(self) -> List[Expression]:
```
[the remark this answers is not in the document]
hsiang-wu: Yes. Fixed.

neuralpp/test/quick_tests/symoblic/expression_test.py
hsiang-wu: yes.

neuralpp/symbolic/expression.py
```diff
@@ -0,0 +1,151 @@
...
+        elif i-1 < arity:
+            arguments = self.arguments
+            arguments[i-1] = new_expression
+            return FunctionApplication(self.function, arguments)
+        else:
+            raise IndexError(f"Out of scope. Function only has arity {arity} but you are setting {i-1}th arguments.")
+
+    def replace(self, from_expression: Expression, to_expression: Expression) -> Expression:
+        # recursively do the replacement
+        new_subexpressions = list(map(
```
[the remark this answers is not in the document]
hsiang-wu: switched to that

neuralpp/symbolic/function.py
hsiang-wu: I think my general objection that i feel it's a bit messy to compare "opeartors". Specifically, in Python, the following test would pass:
```
class OperatorTestClass:
    __match_args__ = ("op")

    def __init__(self, op):
        self._operator = operator

    @property
    def op(self):
        return self._operator


def test_operator_matching():
    e = OperatorTestClass(operator.add)
    e2 = OperatorTestClass(operator.add)
    assert operator.add == operator.add
    assert e.op != operator.add  # a bit counterintuitive here
    assert e.op == e.op
    assert e.op == e2.op

    match e:
        case OperatorTestClass(op=operator.add):
            assert False  # not reachable
        case OperatorTestClass(op=op):  # property is also accepted as __match_args__
            assert op != operator.add
            assert op == e2.op
```

neuralpp/symbolic/expression.py
```diff
@@ -0,0 +1,151 @@
...
+            case _:
+                return False
+
+
+class FunctionApplication(Expression):
+    __match_args__ = ("function", "arguments")
+
+    def __init__(self, func: Expression, args: List[Expression]):
+        """`func` is an expression. Legal options are:
+        1. a Python Callable. E.g.,
```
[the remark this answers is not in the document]
hsiang-wu: Yes. I agree we don't have to pose the requirement here. Maybe that comments should be interpreter-specific and reworded?

neuralpp/symbolic/expression.py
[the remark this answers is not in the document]
CactusWin: I think it depends how strongly we want `Expression` to be an abstract class. Not sure if other people have any strong opinions about it. I think it's fair to have the logic be contained in `Expression`.

neuralpp/symbolic/expression.py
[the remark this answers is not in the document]
hsiang-wu: per discussion with @rodrigodesalvobraz, use the `base_type` design.

hsiang-wu:
Removed all `__init__()` in `Expression` to make it indeed virtual, as @rodrigodesalvobraz pointed out in the comments of project plan.

neuralpp/symbolic/function.py
rodrigodesalvobraz: I think this is just a bug in your test code . Your `__init__` method receives argument `op` but uses `operator` instead.

neuralpp/test/quick_tests/symoblic/expression_test.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # 2. a Function. Here Add is a subclass of Function. All Function has a method lambdify(), which
+    # will be handy for interpretation.
+    func2 = BasicConstant(Add())
+    fa2 = BasicFunctionApplication(func2, [constant_one, constant_two])
+
+    # 3. a Variable. In this case the function is uninterpreted. Even it's named "add", it does not necessarily
+    # have to be a function of addition.
+    func3 = BasicConstant(BasicVariable("add"))
+    fa3 = BasicFunctionApplication(func3, [constant_one, fa2])  # use fa1 here, expression can be recursive
```
rodrigodesalvobraz, line 58: I didn't understand the comment about fa1.

rodrigodesalvobraz (changes requested):

neuralpp/symbolic/expression.py
rodrigodesalvobraz: I was just thinking that. I think it is ok to move to 3.10.

neuralpp/symbolic/expression.py
rodrigodesalvobraz: Yes, I think that's a good idea!

neuralpp/symbolic/expression.py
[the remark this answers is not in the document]
rodrigodesalvobraz: I think either way is ok. The guideline I would follow is: if a method definition is going to apply to *most* Expressions, then it can be placed in Expression as a default. I am not sure if this version here is the latest one, I will go check the latest version and leave a comment there if needed.

neuralpp/symbolic/function.py
hsiang-wu: Yeah you're right.

neuralpp/test/quick_tests/symoblic/expression_test.py
hsiang-wu: it's a typo; should be fa2

neuralpp/symbolic/expression.py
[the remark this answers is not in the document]
hsiang-wu: removed comments

neuralpp/symbolic/function.py
hsiang-wu: removed `function.py`

rodrigodesalvobraz (approved):

PR #9, opened 2022-06-13
added use case of z3's solver and context, particularly their copying

xngw:


neuralpp/test/quick_tests/symoblic/z3_usage_test.py
```diff
@@ -97,3 +98,67 @@ def test_sum():
...
+    "C and L" and "C and not L"; if we can clone the Z3 environment/solver, then we can treat it as immutable.
+    Otherwise, we must treat it as mutable and can only check "C and L", call pop(), and then check "C and not L".
+    """
+    # the use case of using only one Solver()
+    s = Solver()
+    x, y = Ints('x y')
+    constraints = And(x > 2, y < 0)
+    s.add(constraints)
+    literal = x > y
+    s.push()  # create a new scope, so z3 knows where it should pop() to
```
rodrigodesalvobraz, line 116: I am not catching the meaning of "knows where it should pop() to", can you please clarify?

neuralpp/test/quick_tests/symoblic/z3_usage_test.py
```diff
@@ -97,3 +98,67 @@ def test_sum():
...
+    s.add(Not(x > y))
+    assert s.check() == unsat
+
+    # z3 has "context", or state/environment.
+    c = Context()
+    s = Solver(ctx=c)
+    x, y = Ints('x y', ctx=c)
+    constraints = And(x > 2, y < 0)
+    s.add(constraints)
+    s2 = s.translate(c)  # "translate" s with the context c, create a new solver object.
```
rodrigodesalvobraz, line 129: Can you clarify this? What does `translate` do, exactly? I am particularly puzzled by the need to pass `c` as an argument to `translate` because `s` had already been created with `c` as context, so why do we need to pass `c` here as an argument? What if I had passed some other `Context` `c2` as an argument here?

rodrigodesalvobraz (changes requested):
Nice!

neuralpp/test/quick_tests/symoblic/z3_usage_test.py
hsiang-wu: I will add an example of using two different `Context`s before this example.

PR #10, opened 2022-06-14
Minor fixes

hsiang-wu:
Problem: the PR renaming `symoblic` to `symbolic` is landed first, and then the PR adding some changes to `symoblic/z3_usage_test.py` is landed.


hsiang-wu:
Also a small refactor, replace `insubclass` with `isinstance` whenever possible.

horizon-blue (approved):
Good catch :D

PR #11, opened 2022-06-16
Typing for expressions.

hsiang-wu:
1. added a new property type for `Expression`; 
2. for `SymPyExpression`, an additional dict is used to keep track of all sub-literals (variables and function), since a sympy object does not keep track of these;
3. fixed all affected unit tests, and added some new unit tests

### DIFF
--- a/neuralpp/symbolic/basic_expression.py
+++ b/neuralpp/symbolic/basic_expression.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import operator
 import builtins
 from neuralpp.symbolic.expression import Expression, FunctionApplication, Variable, Constant, AtomicExpression, \
-    VariableNotTypedError, FunctionType
+    VariableNotTypedError, ExpressionType
 from abc import ABC
 from typing import Any, List, Optional, Type, Callable, Tuple
 
@@ -13,14 +13,14 @@ class AmbiguousTypeError(TypeError, ValueError):
         super().__init__(f"{python_callable} is ambiguous.")
 
 
-def infer_python_callable_type(python_callable: Callable) -> FunctionType:
+def infer_python_callable_type(python_callable: Callable) -> ExpressionType:
     match python_callable:
         # boolean operation
-        case operator.__and__ | operator.__or__ | operator.__xor__:
+        case operator.and_ | operator.or_ | operator.xor:
             raise AmbiguousTypeError(python_callable)  # this is also ambiguous because the arity could be arbitrary
-            # return FunctionType([bool, bool], bool)
-        case operator.__not__:
-            return FunctionType([bool], bool)
+            # return Callable[[bool, bool], bool]
+        case operator.not_:
+            return Callable[[bool], bool]
         # comparison
         case operator.le | operator.lt | operator.ge | operator.gt | operator.eq:
             raise AmbiguousTypeError(python_callable)
@@ -36,26 +36,21 @@ def infer_python_callable_type(python_callable: Callable) -> FunctionType:
 
 class BasicExpression(Expression, ABC):
     @classmethod
-    def new_constant(cls, value: Any, constant_type: Optional[Expression] = None) -> BasicConstant:
-        return BasicConstant(value, constant_type)
+    def new_constant(cls, value: Any, type_: Optional[ExpressionType] = None) -> BasicConstant:
+        return BasicConstant(value, type_)
 
     @classmethod
-    def new_variable(cls, name: str, variable_type: Expression) -> BasicVariable:
-        return BasicVariable(name, variable_type)
+    def new_variable(cls, name: str, type_: ExpressionType) -> BasicVariable:
+        return BasicVariable(name, type_)
 
     @classmethod
     def new_function_application(cls, function: Expression, arguments: List[Expression]) -> BasicFunctionApplication:
         return BasicFunctionApplication(function, arguments)
 
 
-def new_type(python_type: Type) -> Expression:
-    """ A handy method to create properly wrapped type as an Expression. E.g., new_type(int). """
-    return BasicConstant(python_type)
-
-
 class BasicAtomicExpression(BasicExpression, AtomicExpression, ABC):
     def __init__(self, atom: Any, expression_type: Optional[Expression] = None):
-        if expression_type is None and not Expression.is_internal_type(atom):
+        if expression_type is None and not isinstance(atom, ExpressionType):
             # try to infer type for atom
             if isinstance(atom, Callable):
                 internal_type = infer_python_callable_type(atom)
@@ -71,15 +66,15 @@ class BasicAtomicExpression(BasicExpression, AtomicExpression, ABC):
 
 
 class BasicVariable(BasicAtomicExpression, Variable):
-    def __init__(self, name: str, variable_type: Expression):
-        if variable_type is None:
+    def __init__(self, name: str, type_: Expression):
+        if type_ is None:
             raise VariableNotTypedError
-        BasicAtomicExpression.__init__(self, name, variable_type)
+        BasicAtomicExpression.__init__(self, name, type_)
 
 
 class BasicConstant(BasicAtomicExpression, Constant):
-    def __init__(self, value: Any, constant_type: Optional[Expression] = None):
-        BasicAtomicExpression.__init__(self, value, constant_type)
+    def __init__(self, value: Any, type_: Optional[Expression] = None):
+        BasicAtomicExpression.__init__(self, value, type_)
 
     # just add this one to simplify debugging
     def __str__(self) -> str:
@@ -88,7 +83,7 @@ class BasicConstant(BasicAtomicExpression, Constant):
 
 class BasicFunctionApplication(BasicExpression, FunctionApplication):
     def __init__(self, function: Expression, arguments: List[Expression]):
-        function_type = BasicConstant(function.get_return_type(len(arguments)))
+        function_type = function.get_return_type(len(arguments))
         super().__init__(function_type)
         self._subexpressions = [function] + arguments
 
--- a/neuralpp/symbolic/basic_expression.py
+++ b/neuralpp/symbolic/basic_expression.py
@@ -1,25 +1,68 @@
 from __future__ import annotations
-from neuralpp.symbolic.expression import Expression, FunctionApplication, Variable, Constant, AtomicExpression
+
+import operator
+import builtins
+from neuralpp.symbolic.expression import Expression, FunctionApplication, Variable, Constant, AtomicExpression, \
+    VariableNotTypedError, FunctionType
 from abc import ABC
-from typing import Any, List, Type
+from typing import Any, List, Optional, Type, Callable, Tuple
+
+
+class AmbiguousTypeError(TypeError, ValueError):
+    def __init__(self, python_callable: Callable):
+        super().__init__(f"{python_callable} is ambiguous.")
+
+
+def infer_python_callable_type(python_callable: Callable) -> FunctionType:
+    match python_callable:
+        # boolean operation
+        case operator.__and__ | operator.__or__ | operator.__xor__:
+            raise AmbiguousTypeError(python_callable)  # this is also ambiguous because the arity could be arbitrary
+            # return FunctionType([bool, bool], bool)
+        case operator.__not__:
+            return FunctionType([bool], bool)
+        # comparison
+        case operator.le | operator.lt | operator.ge | operator.gt | operator.eq:
+            raise AmbiguousTypeError(python_callable)
+        # arithmetic
+        case operator.add | operator.mul | operator.pow:
+            raise AmbiguousTypeError(python_callable)
+        # min/max
+        case builtins.min | builtins.max:
+            raise AmbiguousTypeError(python_callable)
+        case _:
+            raise ValueError(f"Python callable {python_callable} is not recognized.")
 
 
 class BasicExpression(Expression, ABC):
     @classmethod
-    def new_constant(cls, value: Any) -> BasicConstant:
-        return BasicConstant(value)
+    def new_constant(cls, value: Any, constant_type: Optional[Expression] = None) -> BasicConstant:
+        return BasicConstant(value, constant_type)
 
     @classmethod
-    def new_variable(cls, name: str) -> BasicVariable:
-        return BasicVariable(name)
+    def new_variable(cls, name: str, variable_type: Expression) -> BasicVariable:
+        return BasicVariable(name, variable_type)
 
     @classmethod
     def new_function_application(cls, function: Expression, arguments: List[Expression]) -> BasicFunctionApplication:
         return BasicFunctionApplication(function, arguments)
 
 
+def new_type(python_type: Type) -> Expression:
+    """ A handy method to create properly wrapped type as an Expression. E.g., new_type(int). """
+    return BasicConstant(python_type)
+
+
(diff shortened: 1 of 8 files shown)


neuralpp/symbolic/expression.py
```diff
@@ -140,13 +204,21 @@ def value(self) -> Any:
...
 
     @property
     @abstractmethod
     def function(self) -> Expression:
         pass
 
+    @property
+    @abstractmethod
+    def number_of_arguments(self) -> int:
+        # in some implementation getting number_of_arguments without calling arguments is useful,
```
rodrigodesalvobraz, line 217: Nice!

neuralpp/symbolic/expression.py
```diff
@@ -74,15 +112,41 @@ def _convert(cls, from_expression: Expression) -> Expression:
...
+    def get_function_type(self) -> FunctionType | Type:
+        """ Call me when self is a function """
+        match self:
+            case Constant(type=type_expression) | Variable(type=type_expression):
+                return type_expression.get_function_type_of_type_expression()
+            case FunctionApplication(function=function, number_of_arguments=num):
+                return function.get_return_type(num)
+
+    def get_function_type_of_type_expression(self) -> FunctionType | Type:
+        """ Call me when self is a type Expression of a function """
```
rodrigodesalvobraz, line 137: Using "call me" for the function documentation (1st person for the function) is clear, but unusual and for that reason may be distracting. Can you please switch to a 3rd person description?

neuralpp/symbolic/sympy_expression.py
```diff
@@ -1,75 +1,121 @@
...
+        case sympy.Float():
+            return BasicConstant(float)
+        case sympy.Rational():
+            return BasicConstant(fractions.Fraction)
+        case sympy.logic.boolalg.BooleanAtom():
+            return BasicConstant(bool)
         case _:
-            raise ValueError(f"Python callable {python_callable} is not recognized.")
+            # It is obvious that we can look up type_dict for symbols like `symbol("x")`.
+            # We can also look up for function, such as `a+b`, where type_dict records the function type (instead of
```
rodrigodesalvobraz, line 31: I didn't quite understand this comment. You write "for function, such as `a+b`", but that is a function application, not a function, so I got confused. It would be helpful to write the comment for `infer_sympy_object_type` telling us what `type_dict` is supposed to contain exactly.
When you say "where type_dict records the function type (instead of the return type)", that sounds like you are defining what `type_dict` is supposed to do/be, in which case that would be better as part of the docstring of the method.
I also got confused by what you meant by "root-level function". Is that the same as "the function used in the function application"? If so, using that would probably be easier to understand.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -1,75 +1,121 @@
...
+            # return type) of the (root-level) function.
+            try:
+                return type_dict[sympy_object]
+            except KeyError:  # if it's not in type_dict, try figure out ourselves
+                return BasicConstant(infer_python_callable_type(sympy_function_to_python_callable(sympy_object)))
+
+
+python_callable_and_sympy_function_pair = [
+    # boolean operation
+    (operator.__and__, sympy.And),
```
rodrigodesalvobraz, line 41: Again here it would be better to use `operator.and` and so on, no?

neuralpp/symbolic/sympy_expression.py
```diff
@@ -1,75 +1,121 @@
...
+    (operator.pow, sympy.Pow),
+    # min/max
+    (builtins.min, sympy.Min),
+    (builtins.max, sympy.Max),
+]
+
+
+def sympy_function_to_python_callable(sympy_function: sympy.Basic) -> Callable:
+    try:
+        return {sympy_function: python_callable
```
rodrigodesalvobraz, line 63: Here you are building this dict every time you look a function up. It seems better to make the dict a property created only the first time it is requested.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -1,75 +1,121 @@
...
+                for python_callable, sympy_function in python_callable_and_sympy_function_pair}[sympy_function]
+    except KeyError:
+        raise ValueError(f"SymPy function {sympy_function} is not recognized.")
+
+
+def python_callable_to_sympy_function(python_callable: Callable) -> sympy.Basic:
+    try:
+        return {python_callable: sympy_function
```
rodrigodesalvobraz, line 71: Same as previous comment.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -1,75 +1,121 @@
...
+            # It is obvious that we can look up type_dict for symbols like `symbol("x")`.
+            # We can also look up for function, such as `a+b`, where type_dict records the function type (instead of
+            # return type) of the (root-level) function.
+            try:
+                return type_dict[sympy_object]
+            except KeyError:  # if it's not in type_dict, try figure out ourselves
+                return BasicConstant(infer_python_callable_type(sympy_function_to_python_callable(sympy_object)))
+
+
+python_callable_and_sympy_function_pair = [
```
rodrigodesalvobraz, line 39: Should the last word in the name be "pairs" rather than "pair"?

rodrigodesalvobraz: Perhaps "relation" (as in math) would be even better.

neuralpp/test/quick_tests/symbolic/expression_test.py
```diff
@@ -16,7 +18,7 @@ def expression_factory(request):
 def test_constant(expression_factory):
     # Constant can be anything
     constant_one = expression_factory.new_constant(1)
-    constant_abc = expression_factory.new_constant("abc")
+    constant_abc = expression_factory.new_constant("abc", BasicConstant(([int], int)))
```
rodrigodesalvobraz, line 21: nit: should this have used `FunctionType`?

neuralpp/symbolic/basic_expression.py
```diff
@@ -1,25 +1,68 @@
...
+    def new_variable(cls, name: str, variable_type: Expression) -> BasicVariable:
+        return BasicVariable(name, variable_type)
 
     @classmethod
     def new_function_application(cls, function: Expression, arguments: List[Expression]) -> BasicFunctionApplication:
         return BasicFunctionApplication(function, arguments)
 
 
+def new_type(python_type: Type) -> Expression:
+    """ A handy method to create properly wrapped type as an Expression. E.g., new_type(int). """
```
rodrigodesalvobraz, line 52: nit: the terms "convenience method" or "helper methods" are more standard.

neuralpp/util/util.py
```diff
@@ -613,3 +613,20 @@ def list_for_each(
...
+
+
+def check_consistency_of_two_dicts(dict1: Dict, dict2: Dict):
+    intersection_keys = dict1.keys() & dict2.keys()
+    for key in intersection_keys:
+        if dict1[key] != dict2[key]:
+            raise ValueError(f"inconsistent dicts on {key}: {dict1[key]} and {dict2[key]}.")
+
+
+def consistently_update_dict(dict1: Dict, dict2: Dict):
```
rodrigodesalvobraz, line 625: nit: would `update_consistent_dicts` be a more precise name? It seems "consistent" is modifying "dicts" rather than "update".

rodrigodesalvobraz (approved):
Awesome, thank you!

neuralpp/symbolic/basic_expression.py
```diff
@@ -1,25 +1,68 @@
...
+            raise AmbiguousTypeError(python_callable)
+        case _:
+            raise ValueError(f"Python callable {python_callable} is not recognized.")
 
 
 class BasicExpression(Expression, ABC):
     @classmethod
-    def new_constant(cls, value: Any) -> BasicConstant:
-        return BasicConstant(value)
+    def new_constant(cls, value: Any, constant_type: Optional[Expression] = None) -> BasicConstant:
```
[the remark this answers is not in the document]
hsiang-wu: I agree. I'm trying to avoid using the word "type" here as it is a keyword (as I try to appease warnings). Any suggestions for the new name? I am thinking of `type_` (in the style of `operator.and_`) or `type` (just ignore warning).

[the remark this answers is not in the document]
rodrigodesalvobraz: `type_` looks good. I've seen `a_type` as well as a solution for this problem. Although here because the function is small, I think `type` is fine too.

neuralpp/symbolic/basic_interpreter.py
```diff
@@ -20,7 +20,7 @@ def eval(self, expression: Expression, context: Expression):
             # there's three cases of `function` that BasicInterpreter can eval():
             # 1. a Python callable, which we'll directly call
             # 2. an uninterpreted function, which is not callable. We raise Error when encounter it.
-            case FunctionApplication(function=Constant(value=python_callable), arguments=args):
+            case FunctionApplication(function=Constant(value=python_callable), arguments=args, type=_):
```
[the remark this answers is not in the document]
hsiang-wu: I'll revert this change. It is unnecessary, since if `type=_` we can just omit it.

We can pattern match any @property or attribute of a class. The `__match_args__` of a class just serves to facilitate [Matching positional attributes](https://peps.python.org/pep-0636/#matching-positional-attributes). So maybe I should also add `type` to `__match_args__`.

neuralpp/symbolic/basic_interpreter.py
```diff
@@ -20,7 +20,7 @@ def eval(self, expression: Expression, context: Expression):
...
                 # * is used to turn a list into "args": https://docs.python.org/2/reference/expressions.html#calls
                 return python_callable(*[self.eval(e, BasicConstant(True)) for e in args])
```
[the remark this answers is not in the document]
hsiang-wu: Yes. :D

neuralpp/symbolic/expression.py
```diff
@@ -1,9 +1,47 @@
...
+
+    @property
+    def return_type(self):
+        return self._return_type
+
+    @property
+    def arity(self):
+        return len(self._argument_types)
+
+    def return_type_after_application(self, number_of_arguments: int) -> FunctionType | Type:
```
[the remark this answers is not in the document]
hsiang-wu: I think there's 2 major reasons why `FunctionType` should not subclass `Type`:

1. There seems to be some strict rule in Python for subclassing `Type` or `type`. In particular, we have no way to initialize a class like the following
```
class Bar(type):
    def __init__(self, foo1, foo2):
        pass
```
In particular, any subclass of `Type`/`type` must have three arguments: a name, a tuple and a dict (haven't looked into the reason, but if you call `Bar("a", "b", {})` you'll get error complaining the second argument is not a tuple. If you do not call with 3 arguments, you'll also get complains.
```
class Bar(Type):
    def __init__(self, name, tuple_, dict_):
        pass

bar = Bar("a", (), {})
```

So we'll need to refactor `FunctionType`'s `__init__` method, at lease.

2. The subclass of typing has the following properties (https://stackoverflow.com/questions/49142505/subclassing-type-vs-object-in-python3):
```
# Its instances are types
MyClass = Bar('MyClass', (), {})

isinstance(MyClass, type) # True

# And it is a class
x = MyClass()

isinstance(x, MyClass) # True
```

Which I believe `FunctionType` should not have. It is more like an annotation of type instead of this usage: when we initialize a `FunctionType` we are creating one particular instance of "function type annotation", not a `class`. I believe it is the same reason why `isinstance(Callable, Type) == False` and `isinstance(List, Type) == False`.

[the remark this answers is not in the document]
hsiang-wu: tldr: `FunctionType` is not `Type`

neuralpp/symbolic/sympy_expression.py
hsiang-wu: yes

neuralpp/symbolic/expression.py
hsiang-wu: yes

neuralpp/symbolic/expression.py
```diff
@@ -74,15 +112,41 @@ def _convert(cls, from_expression: Expression) -> Expression:
...
                 return cls.new_function_application(function, arguments)
             case _:
                 raise ValueError(f"invalid from_expression {from_expression}: {type(from_expression)}")
 
+    @property
+    def type(self) -> Expression:
+        return self._type
+
+    def get_function_type(self) -> FunctionType | Type:
+        """ Call me when self is a function """
```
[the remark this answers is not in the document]
hsiang-wu: I'll no longer have this confusion once we change the type of property `type` from `Expression` to `ExpressionType`

neuralpp/symbolic/expression.py
```diff
@@ -74,15 +112,41 @@ def _convert(cls, from_expression: Expression) -> Expression:
...
+            case Constant(type=type_expression) | Variable(type=type_expression):
+                return type_expression.get_function_type_of_type_expression()
+            case FunctionApplication(function=function, number_of_arguments=num):
+                return function.get_return_type(num)
+
+    def get_function_type_of_type_expression(self) -> FunctionType | Type:
+        """ Call me when self is a type Expression of a function """
+        match self:
+            case Constant(value=internal_type):
+                if not Expression.is_internal_type(internal_type):
```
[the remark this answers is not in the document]
hsiang-wu: changed

neuralpp/symbolic/sympy_expression.py
hsiang-wu: I just created a global one... I don't want to go through the "emulating a static variable of a function" stuff...

neuralpp/symbolic/sympy_expression.py
hsiang-wu: yes

neuralpp/symbolic/basic_expression.py
hsiang-wu: this function is gone now, but noted.

neuralpp/util/util.py
hsiang-wu: yes. thx!

hsiang-wu:
fixes per review

neuralpp/symbolic/basic_interpreter.py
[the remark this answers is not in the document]
rodrigodesalvobraz: I guess no harm leaving it there if it enables matching without specifying attributes.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -7,41 +7,44 @@
...
+            # function type (instead of return type) of the function used in the function application.
             try:
                 return type_dict[sympy_object]
             except KeyError:  # if it's not in type_dict, try figure out ourselves
-                return BasicConstant(infer_python_callable_type(sympy_function_to_python_callable(sympy_object)))
+                return infer_python_callable_type(sympy_function_to_python_callable(sympy_object))
 
 
-python_callable_and_sympy_function_pair = [
+python_callable_and_sympy_function_relations = [
```
rodrigodesalvobraz, line 42: Since a relation is a set of tuples (pairs in this case), this should be in the singular.

rodrigodesalvobraz (approved):
Great, I think this last change really made a lot of things fall neatly into place! :-)

neuralpp/symbolic/sympy_expression.py
hsiang-wu: Ah nice catch!

PR #12, opened 2022-06-17
Z3 Expression

hsiang-wu:
implemented z3 expression and unit tests; fixed a bit previous code to facilitate implicit internal conversion


neuralpp/symbolic/expression.py
```diff
@@ -95,14 +101,19 @@ def _convert(cls, from_expression: Expression) -> Expression:
             return from_expression
         match from_expression:
             case Constant(value=value, type=type_):
-                return cls.new_constant(value, type_)
+                return cls.new_constant(from_expression.pythonize_value(value), type_)
```
rodrigodesalvobraz, line 104: Aren't values of Constants always Python objects? Why do we need to pythonize it?
PS: later in the code I see that SymPy constants may have SymPy objects as values, but I think that is incorrect. The property `value` should provide a Python object for all `Expression` classes. `SymPy.value` should pythonize its SymPy object before returning it.

neuralpp/symbolic/z3_expression.py
```diff
@@ -0,0 +1,335 @@
...
+        case z3.Z3_OP_MUL:
+            return operator.mul
+        case z3.Z3_OP_POWER:
+            return operator.pow
+        case _:
+            raise ValueError(f"Z3 function {z3_function} is not recognized.")
+
+
+def z3_object_to_expression(z3_object: z3.ExprRef) -> Z3Expression:
+    # Here we just try to find a type of expression for z3 object.
```
rodrigodesalvobraz, line 144: I think this comment can be a bit more precise: you are not just trying to find a type, but to build the corresponding `Expression` object. How about "Returns an instance of the appropriate subclass of `Expression` matching `z3_object`"?

neuralpp/symbolic/z3_expression.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+class Z3Expression(Expression, ABC):
+    def __init__(self, z3_object: z3.ExprRef | z3.FuncDeclRef):
+        super().__init__(get_type_from_z3_object(z3_object))
+        self._z3_object = z3_object
+
+    @classmethod
+    def new_constant(cls, value: Any, type_: Optional[ExpressionType] = None) -> Z3Constant:
+        if isinstance(value, z3.ExprRef | z3.FuncDeclRef):
```
rodrigodesalvobraz, line 211: Here we touch on the issue of `value` needing to be a Python object everywhere. We probably need all `Z3Expression` implementations (same for `SymPyExpression` implementations) to contains an attribute `base_object` or some such name of type `z3.ExprRef` | `z3.FunDeclRef` (a sympy object type in the case of `SymPyExpression`) and then a property `value` which returns the corresponding Python object.

Note that the same `base_object` can be used to represent the underlying library (Z3 or SymPy) object even for `Z3FunctionApplication` and `Z3Variable`  (`SymPyFunctionApplication` and `SymPyVariable`), but there it is used to produce the appropriate properties (function, arguments, name, etc)

neuralpp/test/quick_tests/symbolic/expression_test.py
```diff
@@ -208,3 +217,51 @@ def test_sympy_function_application():
...
     assert mixed_adds.subexpressions[2].subexpressions[1].type == float
     assert mixed_adds.subexpressions[2].subexpressions[2].type == float
+
+
+def test_z3_function_application():
+    """
+    We cannot mimic the test above because we do not support z3's float in function yet.
+    (see z3_usage_test.py: test_z3_fp_sort())
+    Also, z3 requires add/compare/... to be of same type, so we cannot have "add: real -> int -> real".
+    In that case explicit conversion is requires (z3.ToInt()/z3.ToReal())
```
rodrigodesalvobraz, line 227: typo: required

rodrigodesalvobraz (changes requested):
Great stuff! However I think we need to enforce the property that `Constant.value` is always a Python (as opposed to Z3 or SymPy value). See comments on review.

neuralpp/symbolic/expression.py
hsiang-wu: fixed

rodrigodesalvobraz (approved):
Great!

PR #13, opened 2022-06-20
Sympy simplification

hsiang-wu:
Implemented a simple `simplify()`;
fixed some nits of misplaced `Expression` where it should be `ExpressionType`.


neuralpp/test/quick_tests/symbolic/interpreter_test.py
```diff
@@ -144,3 +147,25 @@ def test_sympy_interpreter():
                                                                  [SymPyVariable(x, int),
                                                                   SymPyVariable(y, int)])
     assert si.eval(min_of_three_five, dict_to_sympy_context(dict1)) == 3
+
+
+def test_sympy_interpreter_simplify():
```
rodrigodesalvobraz, line 152: It seems we need a few more tests here, particularly involving value bindings, such as the one I mentioned `x + 2` with `x` having value `2`.

rodrigodesalvobraz (changes requested):
Thanks! Not sure we don't need to replace variables by their values *before* simplifying, see comments.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -209,9 +216,18 @@ def atom(self) -> Any:
...
+        Calling by function_type=None asks this function to try to infer the function type.
+        If the caller knows the function_type, it should always set function_type to a non-None value.
+        This function always set type_dict[sympy_object] with the new (inferred or supplied) function_type value.
+        The old value, if exists, is only used for consistency checking.
+        """
         if function_type is None:
-            # it's useless to supply type_dict since we are looking for self.type, which is not set in type_dict yet.
-            function_type = infer_sympy_object_type(sympy_object.func, {})  # almost useless, can only infer "not"
+            # Do not look up this function type from type_dict:
+            # 1. it's useless: in most cases it does not exist, since library calling functions with function_type=None
```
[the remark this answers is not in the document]
hsiang-wu: I want to mean "functions from this library (`neuralpp/symbolic`) that call this function", which I want to distinguish from function calls from "user" (e.g., calling `SymPyFunctionApplication()` in the unit test).

neuralpp/symbolic/sympy_expression.py
```diff
@@ -209,9 +216,18 @@ def atom(self) -> Any:
...
+            #    do not know the function_type.
+            # 2. if the old value exists, it can be erroneous and thus should not be used.
```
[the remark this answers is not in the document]
hsiang-wu: consistent existing types are allowed, otherwise it's an error.

neuralpp/symbolic/sympy_interpreter.py
```diff
@@ -35,17 +35,35 @@ def context_to_variable_value_dict(context: Expression) -> \
...
-    def eval(self, expression: SymPyExpression, context: Expression = BasicConstant(True)):
+    @staticmethod
+    def _simplify_expression(expression: SymPyExpression, context: Expression) -> sympy.Basic:
         variable_value_dict = context_to_variable_value_dict(context)
         # in creation of function application, we set evaluate=False, so 1 + 2 will not evaluate
         # call simplify() evaluates that
         result = expression.sympy_object.simplify()
         for variable, value in variable_value_dict.items():
             result = result.replace(variable, value)
+        return result
```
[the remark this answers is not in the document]
hsiang-wu: `simplify()` is mainly for when `context=Constant(True)`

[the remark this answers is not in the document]
hsiang-wu: switch to only using `simplify()` when `variable_value_dict` is empty

neuralpp/symbolic/sympy_expression.py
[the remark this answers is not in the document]
rodrigodesalvobraz: Can you please modify the comment to make that more clear? Another point that is not clear to me is whether library functions calling this function always do so with `function_type=None` or just sometimes.

neuralpp/test/quick_tests/symbolic/interpreter_test.py
```diff
@@ -62,15 +62,18 @@ def boolean_function_of_arity(arity: int) -> BasicConstant:
 
 
 def dict_to_sympy_context(kv_map: dict) -> SymPyExpression:
-    result = sympy.S.true
+    result = True
```
rodrigodesalvobraz, line 65: The name `result` suggests this is the thing to be returned, but that doesn't seem to be true. Should it really be called `result`?

neuralpp/symbolic/sympy_expression.py
```diff
@@ -185,9 +178,32 @@ def sympy_object(self):
     def type_dict(self) -> Dict[sympy.Basic, Expression]:
         return self._type_dict
 
+    @staticmethod
+    def from_sympy_object(sympy_object: sympy.Basic, argument_type: ExpressionType,
+                          type_dict: Dict[sympy.Basic, Expression]) -> SymPyExpression:
+        # Here we just try to find a type of expression for sympy object.
+        if type(sympy_object) == sympy.Symbol:
+            return SymPyVariable(sympy_object, argument_type)
```
horizon-blue, line 186: Any reason to use `type(...) == ...` instead of `isinstance`? ;)

rodrigodesalvobraz (approved):

neuralpp/symbolic/sympy_interpreter.py
```diff
@@ -35,17 +36,62 @@ def context_to_variable_value_dict(context: Expression) -> \
...
+        # by a context of {y:2}, we'd have
+        #   {x:int, 2:int, 4:int->int->int, x+4:int->int->int} (entry 2 and 4 will be GCed)
+        type_dict = SymPyInterpreter._updated_type_dict_by_context(expression.type_dict, context)
+        type_ = type_dict[result]
+        result_expression = SymPyExpression.from_sympy_object(result, type_, type_dict)
+        # The result keeps the known type information from `expression`. E.g., though (y-y).simplify() = 0, it still
+        # keeps the type of `y`. Say the old `y` is of int type, adding that `0` to another `y` of float type will
+        # cause an error. That's why we need garbage_collect_type_dict() for cleaning the old, unreachable type
+        # information.
+        result_expression.garbage_collect_type_dict()
```
horizon-blue, line 96: Sorry for being late to the party. I have a dumb question regarding `garbage_collect_type_dict` and how it's being used here: IIUC the `Expression` class is designed to be immutable, and in most other methods, we always return a new instance of the expression instead of doing in-place modification on `self`.

I understand the motivation to "prune" unused type dict entries, though I'm wondering if pruning the dict on `self` is slightly violating the immutable design? Happy to hear about your thoughts

neuralpp/symbolic/sympy_expression.py
hsiang-wu: no.. I'll change that in my next PR

hsiang-wu: Not really. Gonna change that in my next PR.

neuralpp/symbolic/sympy_interpreter.py
hsiang-wu: My argument is that

1. `result_expression` created at line #91 is still "temporary" and not visible to the caller, so it's virtually immutable; and
2. immutability in programming languages is more or less an virtualization/illusion (e.g., C++ `const` method can still change the state of the object), especially in python which provides no way to enforce that.

That being said, for this specific example, I can change the design a bit to adhere better to the immutability principle. We can "prune" the `type_dict` first and then return `from_sympy_object()`, essentially swapping line #91 and #96.

PR #14, opened 2022-06-22
Operator Overloading

hsiang-wu:


neuralpp/test/quick_tests/symbolic/expression_test.py
```diff
@@ -266,3 +301,9 @@ def test_sympy_z3_conversion():
     add2 = Z3Expression.new_function_application(real_add_func, [constant_one_third2, two_third])
     assert add2.arguments[1] == add0
     assert add2 == add1
+
+
+def test_sympy_neg_weird():
+    x = SymPyVariable(sympy.symbols("x"), int)
+    neg_x = -x
+    assert neg_x.function.value == operator.mul  # because it is actually "-1 * x"
```
hsiang-wu, line 309: I don't know if that's a desired behavior. Not very straight-forward to an end-user.

neuralpp/test/quick_tests/symbolic/expression_test.py
```diff
@@ -186,17 +185,49 @@ def test_function_application(expression_factory):
...
+    expr = b1 & b2 | ~b3
+    assert expr.function.type == Callable[[bool, bool], bool]  # or
+    if expression_factory != SymPyExpression:
+        assert expr.arguments[0].function.type == Callable[[bool, bool], bool]  # and
+        assert expr.arguments[1].function.type == Callable[[bool], bool]  # not
+        expr1 = b1 & True
+        assert expr1.function.type == Callable[[bool, bool], bool]
+    else:  # sympy changes the argument order
+        assert expr.arguments[0].function.type == Callable[[bool], bool]  # not
+        assert expr.arguments[1].function.type == Callable[[bool, bool], bool]  # and
```
hsiang-wu, line 224: Similar as above, sympy internally change the argument order so it would also be a bit surprising to a user.

rodrigodesalvobraz: Yes, should be fine for now too.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -57,8 +59,8 @@ def infer_sympy_object_type(sympy_object: sympy.Basic, type_dict: Dict[sympy.Bas
     (operator.add, sympy.Add),
     (operator.mul, sympy.Mul),
     (operator.pow, sympy.Pow),
-    (operator.sub, sympy.Lambda((abc.x, abc.y), abc.x-abc.y)),  # "lambda x: (-1)*x"
-    (operator.neg, sympy.Lambda((abc.x,), -abc.x)),  # "lambda x: (-1)*x"
+    (operator.sub, sympy_Sub),
+    (operator.neg, sympy_Neg),  # "lambda x: (-1)*x"
```
horizon-blue, line 63: Perhaps we should move the comments up to line 44 (where you define `sympy_Neg`?)

horizon-blue (commented):
Looks great to me overall. However, in general it will be great if you can write a short description in the PR to explain what the code is about to provide the reviewers with some context.

For example, I was a bit confused about why we decided to map `operator.invert` to `Sympy.Not`/`Z3.NOT`, instead of the more seemly intuitive `operator.not_` as originally written. After read through Python's documentation, I came to realize that there's no `__not__` method on Python object so we have to workaround and use `~` to invoke `__invert__`. This is the type of information that will be very valuable to document somewhere, what do you think? :)

hsiang-wu:
> Looks great to me overall. However, in general it will be great if you can write a short description in the PR to explain what the code is about to provide the reviewers with some context.
> 
> For example, I was a bit confused about why we decided to map `operator.invert` to `Sympy.Not`/`Z3.NOT`, instead of the more seemly intuitive `operator.not_` as originally written. After read through Python's documentation, I came to realize that there's no `__not__` method on Python object so we have to workaround and use `~` to invoke `__invert__`. This is the type of information that will be very valuable to document somewhere, what do you think? :)

Noted, thanks!

neuralpp/symbolic/basic_expression.py
```diff
@@ -19,7 +19,7 @@ def infer_python_callable_type(python_callable: Callable) -> ExpressionType:
         case operator.and_ | operator.or_ | operator.xor:
             raise AmbiguousTypeError(python_callable)  # this is also ambiguous because the arity could be arbitrary
             # return Callable[[bool, bool], bool]
-        case operator.not_:
+        case operator.invert:
```
hsiang-wu, line 22: change to invert (`~`) because python does not provide a `__not__()` overloading

neuralpp/symbolic/basic_expression.py
```diff
@@ -53,10 +53,9 @@ def __init__(self, atom: Any, expression_type: Optional[ExpressionType] = None):
         if expression_type is None and not isinstance(atom, ExpressionType):
             # try to infer type for atom
             if isinstance(atom, Callable):
-                internal_type = infer_python_callable_type(atom)
+                expression_type = infer_python_callable_type(atom)
             else:
-                internal_type = type(atom)
-            expression_type = BasicConstant(internal_type, None)
+                expression_type = type(atom)
```
hsiang-wu, line 58: It was a bug, we no longer need to wrap the type with `Constant()`, doing so causes errors.

neuralpp/symbolic/expression.py
```diff
@@ -1,10 +1,43 @@
...
+# or
+# CASE TWO: `Callable` means "the type of all functions"
+#   >>> isinstance(int->int, Callable)
+#   False
+#   >>> isinstance(lambda (x: int) : int := x, Callable)
+#   True
+#
+# It should be noted that our usage of `Callable` here means the first, i.e., "the type of all function types"
+# And in our code we use `isinstance()` for cases like `isinstance(Callable[[int],int], Callable)`.
 ExpressionType = Callable | Type
```
hsiang-wu, line 41: I got confused by return type `Callable` even though I just wrote it several days ago...So wrote these docs.

neuralpp/symbolic/expression.py
```diff
@@ -140,6 +186,59 @@ def get_return_type(self, number_of_arguments: int) -> ExpressionType:
...
+
+    def __or__(self, other: Any) -> Expression:
+        return self._new_binary_operation(other, operator.or_, Callable[[bool, bool], bool])
+
+    def __ror__(self, other: Any) -> Expression:
+        return self._new_binary_operation(other, operator.or_, Callable[[bool, bool], bool], reverse=True)
+
+    def __invert__(self) -> Expression:
+        return self.new_function_application(self.new_constant(operator.invert, Callable[[bool], bool]), [self])
+
```
hsiang-wu, line 241: The core operator overloading part.

neuralpp/test/quick_tests/symbolic/expression_test.py
```diff
@@ -186,17 +185,49 @@ def test_function_application(expression_factory):
...
+                                                                                              int_to_int_to_int),
                                                            [constant_one, constant_two])
-    assert fa == SymPyExpression.new_function_application(add_func, [constant_one, constant_two])
-    fa6 = fa.set(1, SymPyExpression.new_variable("a", int))
-    assert fa6 == SymPyExpression.new_function_application(add_func, [SymPyExpression.new_variable("a", int),
-                                                                      constant_two])
+    assert fa == expression_factory.new_function_application(add_func, [constant_one, constant_two])
+    fa6 = fa.set(1, expression_factory.new_variable("a", int))
+    assert fa6 == expression_factory.new_function_application(add_func, [expression_factory.new_variable("a", int),
+                                                                         constant_two])
```
hsiang-wu, line 195: We were not using the fixture `expression_factory` in the test, fixed that.

neuralpp/symbolic/z3_expression.py
```diff
@@ -188,6 +196,10 @@ def apply_python_callable_on_z3_arguments(python_callable: Callable,
         # arithmetic
         case operator.add:
             return arguments[0] + arguments[1]
+        case operator.sub:
+            return arguments[0] - arguments[1]
+        case operator.neg:
+            return - arguments[1]
```
CactusWin, line 202: Why is this `return - arguments[1]` and not `return - arguments[0]` or `return - arguments`? What is the first argument (`arguments[0]`)?

hsiang-wu: Nice catch! It's a bug. Fixed and added a test case which would have failed.

PR #17, opened 2022-06-24
Belief propogation

shannyang:


neuralpp/experiments/experimental_inference/belief_propagation.py
```diff
@@ -0,0 +1,38 @@
...
+        return BeliefPropagation.msg_from_variable(query, None, factors)
+
+    @staticmethod
+    def msg_from_factor(factor: Factor, out_var: Variable, factors: List[Factor]):
+        neighbors = [v for v in factor.variables if v != out_var]
+        if not neighbors:
+            return factor
+        assignments = DiscreteVariable.assignments_product_dicts(neighbors)
+        msg = BeliefPropagation.mult_factors([BeliefPropagation.msg_from_variable(v, factor, factors) for v in neighbors])
+        return BeliefPropagation.add_factors([msg.condition(a) * factor.condition(a) for a in assignments])
```
rodrigodesalvobraz, line 32: You don't need to iterate over assignments here. In fact, this may be wasteful because sometimes we can compute  `sum_a f(a)`  without iterating over all `a`'s (for example if `f(a)` is `if a = 1 then 0.1 else 0.2`.

To compute a summation `phi_2(w) = sum_{x,y,z} phi_1(x,y,z,w)` over a factor in a way that leverages the factor's internal representation (without having to know what that internal representation is), you can use `phi_1.sum_out([x,y,z])` instead. This will return factor `phi_2` defined on variable `w`.

In your scenario here you will invoke `sum_out` on the `ProductFactor` you created above.

neuralpp/experiments/experimental_inference/belief_propagation.py
```diff
@@ -0,0 +1,38 @@
...
+    def run(query: Variable, factors: List[Factor]):
+        return BeliefPropagation.msg_from_variable(query, None, factors)
+
+    @staticmethod
+    def msg_from_factor(factor: Factor, out_var: Variable, factors: List[Factor]):
+        neighbors = [v for v in factor.variables if v != out_var]
+        if not neighbors:
+            return factor
+        assignments = DiscreteVariable.assignments_product_dicts(neighbors)
+        msg = BeliefPropagation.mult_factors([BeliefPropagation.msg_from_variable(v, factor, factors) for v in neighbors])
```
rodrigodesalvobraz, line 31: Here you can just create a `ProductFactor` based on the list of factors to be multiplied. You want to create a list with the incoming messages and `factor` itself.

neuralpp/experiments/experimental_inference/belief_propagation.py
```diff
@@ -0,0 +1,38 @@
...
+        return BeliefPropagation.add_factors([msg.condition(a) * factor.condition(a) for a in assignments])
+
+    @staticmethod
+    def msg_from_variable(variable: Variable, out_fac: Optional[Factor], factors: List[Factor]):
+        neighbors = [f for f in factors if variable in f.variables and (out_fac is None or f != out_fac)]
+        msgs_in = [BeliefPropagation.msg_from_factor(f, variable, factors) for f in neighbors]
+        return BeliefPropagation.mult_factors(msgs_in)
```
rodrigodesalvobraz, line 38: Here you just need to return a `ProductFactor` of the incoming messages.

neuralpp/inference/graphical_model/representation/factor/factor.py
```diff
@@ -71,11 +71,20 @@ def __mul__(self, other):
         else:
             return self.mul_by_non_identity(other)
 
+    def __add__(self, other):
```
rodrigodesalvobraz, line 74: We don't need these additions for now.

neuralpp/test/quick_tests/graphical_model/representation/belief_propagation_test.py
```diff
@@ -0,0 +1,37 @@
...
+        return 1.0 if (rain_level + sprinkler_on == wetness) else 0.0
+
+    for log_space in {False, True}:
+        c = IntegerVariable("c", 3)
+        r = IntegerVariable("r", 3)
+        s = IntegerVariable("s", 2)
+        w = IntegerVariable("w", 4)
+
+        model = [
+            PyTorchTableFactor([c], cloudy, log_space=log_space),
```
rodrigodesalvobraz, line 27: Nice creation of the model! Here we don't need to test for both values of log_space, let's just use the default!

neuralpp/test/quick_tests/graphical_model/representation/belief_propagation_test.py
```diff
@@ -0,0 +1,37 @@
...
+            PyTorchTableFactor([c, r], rain, log_space=log_space),
+            PyTorchTableFactor([s], sprinkler, log_space=log_space),
+            PyTorchTableFactor.from_function(
+                [w, r, s], wet_grass, log_space=log_space
+            ),
+        ]
+
+        query = w
+        print(BeliefPropagation().run(query, model)) # [0.192, 0.308, 0.348, 0.152]
```
rodrigodesalvobraz, line 36: It's impressive you got it to work using the additions! That was quite some work. But you will see things will be a lot simpler when you use `sum_out` and `ProductFactor`.

neuralpp/inference/graphical_model/representation/table/pytorch_table.py
```diff
@@ -215,6 +215,37 @@ def __mul__(self, other):
             array_of_potentials = self.potentials_tensor() * other.potentials_tensor()
             return self.from_array(array_of_potentials, result_is_batch)
 
+    def __add__(self, other):
```
rodrigodesalvobraz, line 218: Wow, you really did put in a lot of work into this! But we don't need this new code! :-)

neuralpp/test/quick_tests/graphical_model/representation/belief_propagation_test.py
rodrigodesalvobraz, line 36: Note that, in principle, you might not get a normalized distribution here (for example when you have observations). So you probably want to use `normalize()` at the end of `BeliefPropagation.run`.

BTW, please add a test in which we observe `cloudy` to be `2` (the maximum value). To incorporate observations, condition all factors on dict `{c: 2}`. Do we see an increase in the probability of `wet_grass`? Remember that you will need to include a normalization in the inference algorithm in  order to obtain a normalized distribution that makes more sense.

neuralpp/test/quick_tests/graphical_model/representation/belief_propagation_test.py
```diff
@@ -0,0 +1,37 @@
...
+from neuralpp.inference.graphical_model.representation.factor.pytorch_table_factor import (
+    PyTorchTableFactor,
+)
+from neuralpp.inference.graphical_model.variable.integer_variable import IntegerVariable
+
+
+def test_belief_propagation():
+    cloudy = [0.2, 0.4, 0.4]
+    sprinkler = [0.6, 0.4]
+    rain = [
```
rodrigodesalvobraz, line 11: Note that in our example, `rain` depended on `cloudy`.

rodrigodesalvobraz (changes requested):

neuralpp/test/quick_tests/graphical_model/representation/belief_propagation_test.py
rodrigodesalvobraz, line 11: bit: you might want to rename this variable to reflect the fact that it is the probability of `rain` given `cloudy`. In fact, you might want to name them `prob_...` to differentiate them from the variables themselves.

neuralpp/experiments/experimental_inference/belief_propagation.py
```diff
@@ -0,0 +1,29 @@
...
+class BeliefPropagation:
+
+    @staticmethod
+    def run(query: Variable, factors: List[Factor]):
+        return BeliefPropagation.msg_from_variable(query, None, factors).normalize()
+
+    @staticmethod
+    def msg_from_factor(factor: Factor, out_var: Variable, factors: List[Factor]):
+        neighbors = [v for v in factor.variables if v != out_var]
+        if not neighbors:
```
rodrigodesalvobraz, line 18: I've just created a static method `ProductFactors.multiply(factors)` which returns `ProductFactors(factors)` if there are 0 or more than 1 factors in the list, and `factor[0]` if `len(factors) == 1`. If you update your repo, you can use it here and eliminate the `if not neighbors`.

neuralpp/experiments/experimental_inference/belief_propagation.py
```diff
@@ -0,0 +1,29 @@
+from typing import List, Optional
```
rodrigodesalvobraz, line 1: It looks like you did an amend rather than create a new commit (I only see one). For the future please create a new commit so I can see what the recent changes were specifically.

rodrigodesalvobraz (approved):
Great, looking very nice! :-)

PR #19, opened 2022-06-24
Context class and Z3 Solver Expression

hsiang-wu:


neuralpp/symbolic/expression.py
```diff
@@ -346,6 +353,23 @@ def __str__(self) -> str:
         return f"{self.function}({argument_str})"
 
 
+class Context(Expression, ABC):
```
hsiang-wu, line 356: Added a new subclass of `Expression`: `Context`. It has a `dict` property to provide variable-value lookup, and a `unsatisfiable` property if the condition is unsatisfiable.

neuralpp/symbolic/expression.py
```diff
@@ -346,6 +353,23 @@ def __str__(self) -> str:
...
+    @property
+    @abstractmethod
+    def unsatisfiable(self) -> bool:
+        pass
+
+    @property
+    def and_priority(self) -> int:
```
hsiang-wu, line 363: This is for `Expression & Context` to use `Context`'s `__and__()` instead of `Expression`'s `__and__()` by default.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -314,3 +315,43 @@ def from_sympy_function_and_general_arguments(sympy_function: sympy.Basic, argum
             with sympy.evaluate(False):
                 sympy_object = sympy_function(*[sympy_argument.sympy_object for sympy_argument in sympy_arguments])
                 return SymPyFunctionApplication(sympy_object, type_dict)
+
+
+def context_to_variable_value_dict_helper(context: FunctionApplication,
```
hsiang-wu, line 320: moved from `sympy_interpreter.py`, change the interface from `Dict[sympy.Basic, Any]` to `Dict[str, Any]` which is a unified interface for all contexts.

neuralpp/symbolic/z3_expression.py
```diff
@@ -349,3 +365,118 @@ def subexpressions(self) -> List[Expression]:
...
+    def arguments(self) -> List[Expression]:
+        return [z3_object_to_expression(assertion) for assertion in self.assertions]
+
+    @property
+    def subexpressions(self) -> List[Expression]:
+        return [self.function] + self.arguments
+
+    @property
+    def unsatisfiable(self) -> bool:  # self should always be satisfiable
+        return False
```
hsiang-wu, line 442: We can be sure about this because `Expression` is immutable. Whenever we `make()` an unsatisfiable context, we return `FalseContext()`. So `Z3SolverExpression()` created by our symbolic library is always satisfiable.

Though user needs to follow `__init__()`'s assumption and never passes a `unsat` solver to it. (We can check for `z3_solver.check() == sat` in `__init__()` but that would be some redundant computation to library calls.)

hsiang-wu: On a second thought, it may be better to have this `check() == sat`, since it's good for correctness, and I believe z3 should have some caching mechanism if I `check()` a same solver twice.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -314,3 +315,43 @@ def from_sympy_function_and_general_arguments(sympy_function: sympy.Basic, argum
...
+                variable_to_value = context_to_variable_value_dict_helper(sub_context, variable_to_value)
+        case FunctionApplication(function=Constant(value=operator.eq),
+                                 arguments=[Variable(name=lhs), Constant(value=rhs)]):
+            # the leaf case
+            variable_to_value[lhs] = rhs
+        # all other cases are ignored
+    return variable_to_value
+
+
+def context_to_variable_value_dict(context: FunctionApplication) -> \
```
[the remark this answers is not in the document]
hsiang-wu: Yes and it should never return a `SymPy.Integer`. It's a wrong type annotation. In line 333 we use `value=rhs`, so the value is always a "pythonized" value.

neuralpp/symbolic/z3_expression.py
```diff
@@ -349,3 +365,120 @@ def subexpressions(self) -> List[Expression]:
     @property
     def number_of_arguments(self) -> int:
         return self.z3_object.num_args()
+
+
+z3_false = z3.BoolVal(False)
+
+
+def extract_key_value_from_assertions_helper(assertions: List[z3.ExprRef] | z3.AstVector, dict_: Dict[str, Any]):
```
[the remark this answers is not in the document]
hsiang-wu: Yes

neuralpp/symbolic/sympy_expression.py
```diff
@@ -314,3 +315,43 @@ def from_sympy_function_and_general_arguments(sympy_function: sympy.Basic, argum
...
+        Dict[str, int | sympy.Integer]:
+    return context_to_variable_value_dict_helper(context, {})
+
+
+class SymPyContext(SymPyFunctionApplication, Context):
```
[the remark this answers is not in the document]
hsiang-wu: I'll go ahead with this raise semantics. Please comment on the new commits if you have other thoughts.

neuralpp/test/quick_tests/symbolic/z3_usage_test.py
```diff
@@ -258,3 +258,20 @@ def test_z3_fp_sort():
...
+
+def test_z3_simplify():
+    x = Int('x')
+    g = Goal()
+    g.add(x > 1)
+    g.add(x < 3)
+    r = Tactic('simplify')(g)
+    r = simplify(r.as_expr())
+    # limited, cannot get x == 2
+    assert repr(r) == "And(Not(x <= 1), Not(3 <= x))"
```
CactusWin, line 271: I thought this should be  `"And(Not(x <= 1), Not(3 >= x))"`, since `x < 3` becomes `Not(x >= 3)` and not `Not(x <= 3)`.  Does this test pass?

horizon-blue: @CactusWin isn't `3 <= x` equivalent to `x >= 3` ;)?

neuralpp/util/z3_util.py
```diff
@@ -0,0 +1,17 @@
...
+    """  Make a new solver with constraints from both solver1 and solver2. This function does not modify arguments. """
+    result = copy(solver1)
+    result.append(solver2.assertions())
+    return result
+
+
+def z3_add_solver_and_literal(solver: Solver, constraint: Any) -> Solver:
+    """  Make a new solver from the older solver and the added constraint. This function does not modify arguments. """
+    result = copy(solver)
+    result.add(constraint)
```
horizon-blue, line 16: Fun fact: looks like `solver.append`, `solver.insert`, and `solver.add` are alias of the [same method: `solver. assert_exprs`](https://github.com/Z3Prover/z3/blob/master/src/api/python/z3/z3.py#L7014-L7049)

horizon-blue (approved):
Looks great to me overall!

neuralpp/test/quick_tests/symbolic/z3_usage_test.py
CactusWin: Ignore this! I can't read this morning

neuralpp/symbolic/sympy_expression.py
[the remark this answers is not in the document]
rodrigodesalvobraz: > simplifier where the type of context is SymPyFunctionApplication

Do you mean the SymPy simplifier? Can't you use just TrueContext there?

BTW, I don't know the term "raise semantics", but I agree that if Z3 does it then it is ok for others to do it too. In any case it seems like we don't need SymPyContext, especially if it is just to support old test code.

rodrigodesalvobraz (approved):

PR #15, opened 2022-06-23
Symbolic Factor

CactusWin:
Created a Symbolic Factor class to help integrate in Xiang's work into NeuralPP
- Currently working on cleaning this code up

PR #20, opened 2022-06-28
Symbolic Factor

CactusWin:
Continuation of https://github.com/rodrigodesalvobraz/neuralpp/pull/15
Created a Symbolic Factor class to help integrate in Xiang's work into NeuralPP

### DIFF
--- a/neuralpp/inference/graphical_model/representation/factor/symbolic_factor.py
+++ b/neuralpp/inference/graphical_model/representation/factor/symbolic_factor.py
@@ -1,0 +1,133 @@
+from typing import List, Any, Optional, Type, Callable, Dict
+import operator
+
+from neuralpp.inference.graphical_model.representation.factor.atomic_factor import AtomicFactor
+from neuralpp.inference.graphical_model.variable.discrete_variable import DiscreteVariable
+from neuralpp.inference.graphical_model.variable.variable import Variable
+from neuralpp.inference.graphical_model.variable.integer_variable import IntegerVariable
+from neuralpp.symbolic.basic_expression import BasicConstant, BasicVariable, BasicFunctionApplication
+from neuralpp.symbolic.expression import Expression, FunctionApplication
+from neuralpp.symbolic.sympy_interpreter import SymPyInterpreter
+from neuralpp.util.util import union, join
+
+class SymbolicFactor(AtomicFactor):
+    def __init__(self, variables: List[Variable], expression: Expression):
+        super().__init__(variables)
+        self.expression = expression
+
+    def new_instance(self, variables: List[Variable], expression: Expression):
+        return type(self)(variables, expression)
+
+    def assignments(self):
+        return DiscreteVariable.assignments_product(self.variables)
+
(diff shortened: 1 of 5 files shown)


neuralpp/test/quick_tests/graphical_model/representation/factor/symbolic_factor_test.py
```diff
@@ -0,0 +1,63 @@
...
+from neuralpp.inference.graphical_model.variable.integer_variable import IntegerVariable
+
+int_to_int_to_int = Callable[[int, int], int]
+
+def test_sympy_condition():
+    x = IntegerVariable("x", 3)
+    y = IntegerVariable("y", 2)
+
+    x_sympy, y_sympy = sympy.symbols("x y")
+    expression1 = SymPyFunctionApplication(x_sympy * y_sympy, {x_sympy: int, y_sympy: int})
```
CactusWin, line 16: > You should be able to use `expression1 = x_sympy * y_sympy` here since there is operator overloading.
I don't believe I can, since ` x_sympy` and `y_sympy` are both sympy variables, and not `expressions`. I could make them as expressions and then use operator overloading if that's better.

neuralpp/inference/graphical_model/representation/factor/symbolic_factor.py
```diff
@@ -0,0 +1,133 @@
...
+            return int
+        else:
+            return float
+
+    def _dict_to_context(self, assignment_dict: dict) -> Expression:
+        result = BasicConstant(True)
+        type_dict = {}
+        for k, v in assignment_dict.items():
+            symbol = BasicVariable(k.name, self._variable_type(v))
+            eq_operator = BasicConstant(operator.eq, Callable[[type(v), type(v)], bool])
```
CactusWin, line 35: > Overloaded operator here too.

I'm not quite sure how I can overload `eq` here, since I would need something to contain the expression `symbol = v`. I don't think I can contain that in `BasicVariable` or `BasicConstant`

neuralpp/inference/graphical_model/representation/factor/symbolic_factor.py
```diff
@@ -0,0 +1,133 @@
...
+            eq_expression = BasicFunctionApplication(eq_operator, [symbol, BasicConstant(v, type(v))])
+
+            and_operator = BasicConstant(operator.and_, Callable[[bool, bool], bool])
+            result = BasicFunctionApplication(and_operator, [result, eq_expression])
```
CactusWin, line 39: Interesting note about this is that: if I use the overloaded operator `and`, the answer is incorrect. For example, if I call `condition` on `x * y` with the dictionary `{x: 1, y: 2}`, the `condition` function with the overloaded operator will return `2 * y`, rather than 2

CactusWin: I had to use the `&` operator and not `and`

CactusWin:
TODO: change the implementation to use `Context`

rodrigodesalvobraz (approved):
Looks great!

PR #22, opened 2022-06-29
Conditional expression

hsiang-wu:
1. Added conditional expression; created `funtions.py` that currently contains a `cond()` function; added support for conditional expression in sympy and z3; SymPy uses `PieceWise` which is bit different from `if-then-else` and requires special treatment.
2. Add support for overloading `<`, `>`, `<=`, `>=` and `!=` to create symbolic expression.
3. Changed `__eq__` semantics to creating symbolic expression. The old semantics is renamed to `syntactic_eq()`.

### DIFF
--- a/neuralpp/symbolic/constants.py
+++ b/neuralpp/symbolic/constants.py
@@ -1,0 +1,105 @@
+import fractions
+import operator
+from typing import Callable, Type
+from .basic_expression import BasicConstant
+from .functions import conditional
+
+
+def if_then_else(type_: Type):
+    return BasicConstant(conditional, Callable[[bool, type_, type_], type_])
+
+
+bool_if_then_else = if_then_else(bool)
+int_if_then_else = if_then_else(int)
+float_if_then_else = if_then_else(float)
+real_if_then_else = if_then_else(fractions.Fraction)
+
+
+def add(type_: Type):
+    return BasicConstant(operator.add, Callable[[type_, type_], type_])
+
+
+int_add = add(int)
+float_add = add(float)
+real_add = add(fractions.Fraction)
+
+
+def multiply(type_: Type):
+    return BasicConstant(operator.mul, Callable[[type_, type_], type_])
+
+
+int_multiply = multiply(int)
+float_multiply = multiply(float)
+real_multiply = multiply(fractions.Fraction)
+
+
+def minus(type_: Type):
+    return BasicConstant(operator.sub, Callable[[type_, type_], type_])
+
+
+int_minus = minus(int)
+float_minus = minus(float)
+real_minus = minus(fractions.Fraction)
+
+
+def div(type_: Type):
+    return BasicConstant(operator.truediv, Callable[[type_, type_], type_])
+
+
+int_div = div(int)
+float_div = div(float)
+real_div = div(fractions.Fraction)
+
+
+def lt(type_: Type):
+    return BasicConstant(operator.lt, Callable[[type_, type_], bool])
+
+
+int_lt = lt(int)
+float_lt = lt(float)
+real_lt = lt(fractions.Fraction)
+
+
+def le(type_: Type):
+    return BasicConstant(operator.le, Callable[[type_, type_], bool])
+
+
+int_le = le(int)
+float_le = le(float)
+real_le = le(fractions.Fraction)
+
+
+def gt(type_: Type):
+    return BasicConstant(operator.gt, Callable[[type_, type_], bool])
+
+
+int_gt = gt(int)
+float_gt = gt(float)
+real_gt = gt(fractions.Fraction)
+
+
+def ge(type_: Type):
+    return BasicConstant(operator.ge, Callable[[type_, type_], bool])
+
+
+int_ge = ge(int)
+float_ge = ge(float)
+real_ge = ge(fractions.Fraction)
+
+
+def eq(type_: Type):
+    return BasicConstant(operator.eq, Callable[[type_, type_], bool])
+
+
+int_eq = eq(int)
+float_eq = eq(float)
+real_eq = eq(fractions.Fraction)
+
+
+def ne(type_: Type):
+    return BasicConstant(operator.ne, Callable[[type_, type_], bool])
+
+
+int_ne = ne(int)
+float_ne = ne(float)
+real_ne = ne(fractions.Fraction)
--- a/neuralpp/symbolic/constants.py
+++ b/neuralpp/symbolic/constants.py
@@ -3,9 +3,10 @@ import operator
 from typing import Callable, Type, Any
 from .basic_expression import BasicConstant
 from .functions import conditional
+from .expression import Expression
 
 
-def if_then_else_function(type_: Type):
+def if_then_else_function(type_: Type) -> Expression:
     return BasicConstant(conditional, Callable[[bool, type_, type_], type_])
 
 
@@ -15,7 +16,7 @@ float_if_then_else_function = if_then_else_function(float)
 real_if_then_else_function = if_then_else_function(fractions.Fraction)
 
 
-def if_then_else(if_: Any, then_: Any, else_: Any):
+def if_then_else(if_: Any, then_: Any, else_: Any) -> Expression:
     if not type(then_) == type(else_):
         raise TypeError(f"Expect then-clause ({type(then_)}) and else-clause ({type(else_)}) have the same type.")
     match then_:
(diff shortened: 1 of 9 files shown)


neuralpp/symbolic/expression.py
```diff
@@ -256,20 +272,38 @@ def __and__(self, other: Any) -> Expression:
...
+        return self._new_binary_boolean(other, operator.or_)
 
     def __ror__(self, other: Any) -> Expression:
-        return self._new_binary_operation(other, operator.or_, Callable[[bool, bool], bool], reverse=True)
+        return self._new_binary_boolean(other, operator.or_, reverse=True)
 
     def __invert__(self) -> Expression:
         return self.new_function_application(self.new_constant(operator.invert, Callable[[bool], bool]), [self])
 
+    def __lt__(self, other) -> Expression:
```
hsiang-wu, line 289: new operator overloading

neuralpp/symbolic/sympy_expression.py
```diff
@@ -312,11 +338,63 @@ def from_sympy_function_and_general_arguments(sympy_function: sympy.Basic, argum
             # see test/sympy_test.py: test_sympy_bug()
             sympy_object = sympy_function(*[sympy_argument.sympy_object for sympy_argument in sympy_arguments],
                                           evaluate=False)
-            return SymPyFunctionApplication(sympy_object, type_dict)
+        elif sympy_function == sympy.Piecewise:
+            sympy_object = sympy_piecewise_from_if_then_else(
```
rodrigodesalvobraz, line 342: How do we know `len(arguments) == 3` here? Can't we call this function with more arguments than 3?

neuralpp/symbolic/sympy_expression.py
```diff
@@ -264,7 +268,38 @@ def atom(self) -> Any:
         return SymPyExpression.pythonize_value(self._sympy_object)
 
 
-class SymPyFunctionApplication(SymPyExpression, FunctionApplication):
+class SymPyFunctionApplicationInterface(SymPyExpression, FunctionApplication, ABC):
```
hsiang-wu, line 271: I created an interface because `SymPyCondFunctionApplication` is created for piecewise/cond operators specifically, which overloads `arguments`. The idea is we want a unified "if-then-else" conditional expression interface for all expressions (`expr.arguments[0]` would be the if-clause, `expr.argument[1]` the then-clause, `expr.argument[2]` the else-clause), but internal sympy uses `piecewise` (whose `args[0]` is `(then-clause, if-clause)`), so we need to do something under the hood to unify it to the interface (see `sympy_piecewise_to_if_then_else` and `sympy_piecewise_from_if_then_else` for detail).

neuralpp/symbolic/sympy_expression.py
```diff
@@ -312,11 +338,63 @@ def from_sympy_function_and_general_arguments(sympy_function: sympy.Basic, argum
...
+
+def sympy_piecewise_to_if_then_else(piecewise: sympy.Piecewise) -> Tuple[sympy.Basic, sympy.Basic, sympy.Basic]:
+    return piecewise.args[0][1], piecewise.args[0][0], piecewise.args[1][0]
+
+
+def fold_sympy_piecewise(piecewise_args: List[Tuple[sympy.Basic, sympy.Basic]]) -> sympy.Piecewise:
+    """ `fold` any sympy piecewise to 2-entry piecewise. E.g.,
+    Piecewise((s0, c0), (s1, c1), (s2, True)) will be folded into
+    Piecewise((s0, c0), (Piecewise((s1, c1), (s2, True)), True)
+    Note the last condition must be True, otherwise it cannot be transformed into if-then-else
```
rodrigodesalvobraz, line 363: This line seems out of place, since this function does not regard if-then-else, only piecewise. It seems that for the purposes of this method, that requirement is not needed. Perhaps it is a requirement that needs to be respected in some other part of the code that involves if-then-else?

neuralpp/symbolic/sympy_expression.py
```diff
@@ -312,11 +338,63 @@ def from_sympy_function_and_general_arguments(sympy_function: sympy.Basic, argum
...
+        raise TypeError("piecewise is expected to have at least two entries")
+    elif len(piecewise_args) == 2:
+        if not piecewise_args[1][1]:
+            raise TypeError("No else clause: missing default value.")
+        return sympy.Piecewise(*piecewise_args)
+    else:
+        return sympy.Piecewise(piecewise_args[0], (fold_sympy_piecewise(piecewise_args[1:]), True))
+
+
+class SymPyCondFunctionApplication(SymPyFunctionApplicationInterface):
```
rodrigodesalvobraz, line 375: Please use "Conditional" instead of "Cond" (the idea being that we can always count on identifiers being formed out of full words).

neuralpp/test/quick_tests/symbolic/interpreter_test.py
```diff
@@ -223,14 +226,31 @@ def test_sympy_interpreter_simplify_operator_overload():
...
     # with a context
     y_2_context = dict_to_sympy_context({"y": 2})
     print(f"y=2 context:{y_2_context}")
     assert si.simplify(b_x_plus_y, y_2_context).sympy_object == x + 2
 
+    # if then else:
+    # convert from basic
+    x_2_y_2_context = dict_to_sympy_context({"x": 2, "y": 2})
+    x_3_y_2_context = dict_to_sympy_context({"x": 3, "y": 2})
+    f = BasicConstant(functions.cond, Callable[[bool, int, int], int])
```
rodrigodesalvobraz, line 241: Same comment here regarding "cond"/"conditional".

rodrigodesalvobraz, line 241: Can you please create a function `if_then_else` to make the creation of such expressions easier?

rodrigodesalvobraz (changes requested):
Great!

neuralpp/symbolic/sympy_expression.py
hsiang-wu: No. Because we are always converting from other *if-then-else* style conditional expressions (`functions.cond` or `z3.If()`. So the length of arguments must be 3.

neuralpp/symbolic/sympy_expression.py
hsiang-wu: That makes sense. I'll move the check outside this function.

neuralpp/test/quick_tests/symbolic/interpreter_test.py
hsiang-wu: That makes sense. I think I should create a new PR for a library that contains these functions and maybe even overload the `__call__` operator.

hsiang-wu: Creating a new PR might be too slow and blocking. I just added the `constants.py` (the module name has been discussed with @rodrigodesalvobraz) in the following commit. I also overloaded `__call__` in `Expression` class to create new `FunctionApplication` object.

neuralpp/test/quick_tests/symbolic/expression_test.py
```diff
@@ -231,6 +232,26 @@ def test_function_application(expression_factory):
             expr1 = b1 & True
 
 
+def test_constants_operators():
```
hsiang-wu, line 235: Some usage examples showing that `constants.py` could be handy.

neuralpp/symbolic/constants.py
```diff
@@ -0,0 +1,105 @@
+import fractions
+import operator
+from typing import Callable, Type
+from .basic_expression import BasicConstant
+from .functions import conditional
+
+
+def if_then_else(type_: Type):
```
hsiang-wu, line 8: We need the type information, so we cannot have an `if_then_else` constant.

hsiang-wu:
Added `constants.py`; overloaded `__call__()` in `Expression` to create a new `FunctionApplication` object.

neuralpp/symbolic/constants.py
```diff
@@ -15,7 +16,7 @@ def if_then_else_function(type_: Type):
 real_if_then_else_function = if_then_else_function(fractions.Fraction)
 
 
-def if_then_else(if_: Any, then_: Any, else_: Any):
+def if_then_else(if_: Any, then_: Any, else_: Any) -> Expression:
```
hsiang-wu, line 19: Added return types of the functions in this file; nit found by @horizon-blue

rodrigodesalvobraz (approved):
Great!

PR #23, opened 2022-06-29
Exact belief propagation

shannyang:


neuralpp/experiments/experimental_inference/exact_belief_propagation.py
```diff
@@ -0,0 +1,37 @@
...
+        messages = [self.msg_from_node(v) for v in self.tree.neighbors(node)]
+        if isinstance(node, Variable):
+            self.process_variable_sum_point(node)
+            if not messages:
+                return Group.identity
+        else:
+            messages += [node]
+        return ProductFactor.multiply(messages).atomic_factor().sum_out_variables(self.sum_points.get(node, []))
+
+    def process_variable_sum_point(self, variable: Variable):
```
rodrigodesalvobraz, line 31: Looking at the code here I understand better how you thought about the issue. I am glad you figured this out! However, I think we can make the code significantly simpler. In particular, this function is accessing `tree.graph`, which I think is not ideal because it assumes a certain internal structure for the spanning tree/graph. Ideally, the algorithm should receive _any_ graph and work on that, without using its internal structure or assuming how it was generated. I think this is a bit trickier and it would be good for us to have a meeting before proceeding.

rodrigodesalvobraz (commented):
Hi Shannon! Great work with figuring out the need to decide sum points for each variable, which I had not thought about when we planned this! Right now I think the solution works but is a bit more complicated than we can hope for. Let's have a meeting to design a simpler solution.

neuralpp/test/quick_tests/graphical_model/representation/belief_propagation_test.py
```diff
@@ -30,12 +31,56 @@ def prob_wet_grass(wetness: int, rain_level: int, sprinkler_on: int):
...
     conditioned_model = [f.condition(observations) for f in model]
 
     # this should result in increased chances of rain
     expected_w_with_conditions = PyTorchTableFactor([w], [0.12, 0.26, 0.42, 0.2])
-    assert(BeliefPropagation().run(w, conditioned_model) == expected_w_with_conditions)
+    assert(ExactBeliefPropagation(conditioned_model, True).run(w) == expected_w_with_conditions)
+    assert (ExactBeliefPropagation(conditioned_model, False).run(w) == expected_w_with_conditions)
+
+
+def test_ebp_with_loop():
```
rodrigodesalvobraz, line 47: It would be good to keep the old tree-based test for Belief Propagation (just for completeness).

shannyang:
Addressed some smaller points on names/comments, no need to re-review yet!

neuralpp/test/quick_tests/graphical_model/representation/belief_propagation_test.py
rodrigodesalvobraz: Oh, sorry, I missed that.

rodrigodesalvobraz (changes requested):
Looking great! Just a couple of changes.

neuralpp/experiments/experimental_inference/graph_analysis.py
```diff
@@ -0,0 +1,102 @@
...
+def node_variables(node):
+    return node.variables if isinstance(node, Factor) else [node]
+
+
+class FactorPartialSpanningTree(PartialSpanningTree):
+
+    def __init__(self, graph: FactorGraph, root):
+        super().__init__(graph, root)
+
+        def _evaluate(node):
```
rodrigodesalvobraz, line 62: It is looking to me that `_evaluate` is unnecessary. If you don't do this, when `children` will still be automatically computed when that method is used. This will make the code simpler and potentially improve performance since the calculation only actually happens if needed.

neuralpp/experiments/experimental_inference/graph_analysis.py
```diff
@@ -0,0 +1,102 @@
...
+            for n in self.children(node):
+                _evaluate(n)
+        self._parents[id(root)] = None
+        _evaluate(root)  # Currently evaluates a full spanning tree
+
+    def variables(self, node) -> Iterable[Variable]:
```
rodrigodesalvobraz, line 68: We need to cache these methods (`variables`, `siblings_variables`, `external_variables`) to avoid them being run multiple times (in BeliefPropagation an ancestor's `external_variables` would be recomputed every time any of its descendants `external_variables` is requested).

neuralpp/experiments/experimental_inference/exact_belief_propagation.py
```diff
@@ -0,0 +1,41 @@
...
+    def factor_at(node):
+        return [] if isinstance(node, Variable) else [node]
+
+    def variables_summed_out_at(self, node, incoming_variables):
+        return util.subtract(
+            incoming_variables,
+            self.tree.external_variables(node) | self.tree.variables_in_node_and_ancestors(node))
+
+    def message_from(self, node):
+        product_at_node = self.product_at(node).atomic_factor()
```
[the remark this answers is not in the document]
shannyang: Fixed, it seems that the flatten util was transforming to lists before rather than flattening into the result!

shannyang:
It seems my caching introduced a bug which I only noticed when running the random test repeatedly. I checked that all the commits before this seem fine when run on 10000 random tests.

To do tomorrow:
- [ ] save some of the fail cases and look more into this
- [ ] number of random trials should be increased to make sure this gets caught. Around 100 was enough for this particular incident without really slowing down test runs.

shannyang:
> It seems my caching introduced a bug which I only noticed when running the random test repeatedly. I checked that all the commits before this seem fine when run on 10000 random tests.
> 
> To do tomorrow:
> 
> * [x]  save some of the fail cases and look more into this
> * [ ]  number of random trials should be increased to make sure this gets caught. Around 100 was enough for this particular incident without really slowing down test runs.

Debugged this, it seems like there were cache collisions when two objects had the same `id` (since uniqueness is only guaranteed for objects with overlapping lifetimes). This generally doesn't happen when repeatedly running the test with a trial size of 10, but when running 500+ trials at once it happens quite frequently.

For now it's solved by having the cache add `hash(str(arg))` to the `id` and giving the cache a maximum size. Theoretically this could cause collisions, but so far so good when running 100000 trials. We could also use `str(id) + str(arg)` as the key, for more collision resistance.

This should be fine for review now (though no rush), would like to get thoughts on the caching strategy!

neuralpp/util/cache_by_id.py
```diff
@@ -0,0 +1,30 @@
...
+from typing import Optional
+
+NO_CACHED_VALUE = object()
+
+
+def cache_by_id(fnc):
+    return lru_cache_by_id(None)(fnc)
+
+
+def lru_cache_by_id(capacity: Optional[int]):
```
rodrigodesalvobraz, line 11: Nice!

neuralpp/util/util.py
```diff
@@ -432,7 +432,13 @@ def flatten_one_level(iterable, is_nested, get_nested):
     """
     Returns [get_nested(e) if is_nested(e) else e for e in iterable]
     """
-    return [get_nested(e) if is_nested(e) else e for e in iterable]
+    result = []
```
rodrigodesalvobraz, line 435: It seems you fixed a bug here? I am surprised it was working before... Anyway, the docstring is now not matching the semantics of the method, so can you please change it to reflect what the method is doing?

neuralpp/experiments/experimental_inference/graph_analysis.py
```diff
@@ -32,39 +33,40 @@ def neighbors(self, node):
...
     def __init__(self, graph: FactorGraph, root):
         super().__init__(graph, root)
-
-        def _evaluate(node):
-            for n in self.children(node):
-                _evaluate(n)
         self._parents[id(root)] = None
-        _evaluate(root)  # Currently evaluates a full spanning tree
 
+    @lru_cache_by_id(1000)
```
rodrigodesalvobraz, line 69: I am not sure we want a limit here (1000). We might want to store as much as memory will allow to tradeoff memory for speed. If we are going to allow a limit, I think it is better to let that be a parameter of the class (but I am not sure decorators can use those).

rodrigodesalvobraz (changes requested):
Nice changes! Just a few requests. Thanks!

neuralpp/test/quick_tests/util/caching_test.py
```diff
@@ -0,0 +1,85 @@
...
+class MethodContainer:
+    call_count = 0
+
+    @cache_by_id
+    def add(self, x: int, y: int):
+        self.call_count += 1
+        return x + y
+
+    @cache_by_id
+    def fibonacci(self, n: NumWrapper):
```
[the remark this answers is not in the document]
shannyang: Two separate instances of `NumWrapper(2)` are different objects and therefore should be treated differently under `cache_by_id`, but smaller integers are cached by Python so `2` would always have the same id as any other `2`.

neuralpp/test/quick_tests/util/caching_test.py
```diff
@@ -0,0 +1,85 @@
...
+        self.call_count += 1
+        value = n.value
+        if value <= 1:
+            return 1
+        return self.fibonacci(num_wrappers[value-1]) + self.fibonacci(num_wrappers[value-2])
+
+
+def test_cache_by_id_on_loop():
```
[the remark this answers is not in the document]
shannyang: When the call is executed the count is updated, but

NumWrapper isn't used in the add example at all, I should probably clarify that in the test class.

rodrigodesalvobraz (approved):
Great! :-)

PR #24, opened 2022-07-05
context simplifer

hsiang-wu:
- Define a `ContextSimplifier` subclass of `Simplifier` where the `simplify` method does the following:
  * Apply `SymPyInterpreter.simplify()` to the whole expression (to eliminate as much of the expression as possible)
  * Suppose there is a total order O on variables and constants where constants come before variables. Determine equivalence classes of variables and constants that are equaled by the context. Replace each variable by the minimum of its equivalent class according to O.
  * For every atom Gamma in the expression, replace Gamma by True if Gamma is implied by the context and by False if its negation is implied by the context.
  * Apply `SymPyInterpreter.simplify()` again to the whole expression to take those substitutions into account.

### DIFF
--- a/neuralpp/symbolic/constants.py
+++ b/neuralpp/symbolic/constants.py
@@ -1,9 +1,16 @@
 import fractions
 import operator
 from typing import Callable, Type, Any
+
+import z3
+import sympy
+
+from .expression import Expression
 from .basic_expression import BasicConstant
 from .functions import conditional
 from .expression import Expression
+from .z3_expression import Z3Constant
+from .sympy_expression import SymPyConstant
 
 
 def if_then_else_function(type_: Type) -> Expression:
@@ -16,20 +23,18 @@ float_if_then_else_function = if_then_else_function(float)
 real_if_then_else_function = if_then_else_function(fractions.Fraction)
 
 
-def if_then_else(if_: Any, then_: Any, else_: Any) -> Expression:
-    if not type(then_) == type(else_):
-        raise TypeError(f"Expect then-clause ({type(then_)}) and else-clause ({type(else_)}) have the same type.")
-    match then_:
-        case bool():
-            return bool_if_then_else_function(if_, then_, else_)
-        case int():
-            return int_if_then_else_function(if_, then_, else_)
-        case float():
-            return float_if_then_else_function(if_, then_, else_)
-        case fractions.Fraction():
-            return real_if_then_else_function(if_, then_, else_)
-        case _:
-            raise TypeError(f"Unrecognized type {type(then_)}.")
+def if_then_else(if_: Any, then_: Any, else_: Any):
+    def _type_of(expression: Any):
+        match expression:
+            case Expression(type=type_):
+                return type_
+            case _:
+                return type(expression)
+    then_type = _type_of(then_)
+    else_type = _type_of(else_)
+    if then_type != else_type:
+        raise TypeError(f"Expect then-clause ({then_type}) and else-clause ({else_type}) have the same type.")
+    return if_then_else_function(then_type)(if_, then_, else_)
 
 
 def add(type_: Type) -> Expression:
@@ -120,3 +125,11 @@ def ne(type_: Type) -> Expression:
 int_ne = ne(int)
 float_ne = ne(float)
 real_ne = ne(fractions.Fraction)
+
+
+basic_true = BasicConstant(True, bool)
+basic_false = BasicConstant(False, bool)
+z3_true = Z3Constant(z3.BoolVal(True))
+z3_false = Z3Constant(z3.BoolVal(False))
+sympy_true = SymPyConstant(sympy.S.true, bool)
+sympy_false = SymPyConstant(sympy.S.false, bool)
--- a/neuralpp/symbolic/context_simplifier.py
+++ b/neuralpp/symbolic/context_simplifier.py
@@ -1,0 +1,97 @@
+from typing import Optional, Callable, Set
+from .expression import Expression, Context
+from .simplifier import Simplifier
+from .sympy_interpreter import SymPyInterpreter
+from .constants import basic_true, basic_false
+from .z3_expression import Z3SolverExpression
+from .parameters import sympy_evaluate
+
+
+def _implies(context: Context, expression: Expression) -> bool:
+    """
+    context implies expression iff (context => expression) is valid;
+    which means not (context => expression) is unsatisfiable;
+    which means not (not context or expression) is unsatisfiable;
+    which means context and not expression is unsatisfiable.
+    """
+    new_context = context & ~expression
+    if new_context.satisfiability_is_known:
+        return new_context.unsatisfiable
+    else:
+        return False
+
+
+def _simplify_expression(boolean_expression: Expression, context: Z3SolverExpression) -> Optional[Expression]:
+    if boolean_expression.type != bool:
+        raise TypeError("Can only simplify booleans")
+    if _implies(context, boolean_expression):
+        return basic_true
+    if _implies(context, ~boolean_expression):
+        return basic_false
+    return None
+
+
+def _collect_subset_expressions_helper(expression: Expression, test: Callable[[Expression], bool],
+                                       result: Set[Expression]):
+    if test(expression):
+        result.add(expression)
+    for sub_expression in expression.subexpressions:
+        _collect_subset_expressions_helper(sub_expression, test, result)
+
+
+def _collect_subset_expressions(expression: Expression,
+                                test: Callable[[Expression], bool] = lambda _: True, ) -> Set[Expression]:
+    """
+    Collect all subexpressions including the ones that are not immediate subexpressions and self, with a filter
+    `test`.
+    """
+    result = set()
+    _collect_subset_expressions_helper(expression, test, result)
+    return result
+
+
+class ContextSimplifier(Simplifier):
+    sympy_interpreter = SymPyInterpreter()
+
+    def __init__(self):
+        pass
+
+    @staticmethod
+    def _simplify_pass(expression: Expression, context: Z3SolverExpression) -> Expression:
+        result = ContextSimplifier.sympy_interpreter.simplify(expression, context)
+        assert result is not None
+
+        # replace boolean expressions
+        all_boolean_subexpressions = _collect_subset_expressions(result, lambda expr: expr.type == bool)
+        if result is None:
+            raise
+        for boolean_subexpression in all_boolean_subexpressions:
+            simplified_subexpression = _simplify_expression(boolean_subexpression, context)
+            if simplified_subexpression is not None:
+                result = result.replace(boolean_subexpression, simplified_subexpression)
+                if result is None:
+                    raise
+
+        # replace variables
+        for variable, replacement in context.variable_replacement_dict.items():
+            result = result.replace(variable, replacement)
+        return result
+
+    def simplify(self, expression: Expression, context: Z3SolverExpression) -> Expression:
+        with sympy_evaluate(True):
+            if not isinstance(context, Z3SolverExpression):
+                raise ValueError("ContextSimplifier expects a Z3SolverExpression context.")
+            if not context.satisfiability_is_known:
+                raise Context.UnknownError()
+            if context.unsatisfiable:
+                raise ValueError(f"Context {context} is unsatisfiable.")
+
+            new_expression = ContextSimplifier._simplify_pass(expression, context)
+            # debug_i = 0
+            while not new_expression.syntactic_eq(expression):
+                expression = new_expression
+                new_expression = ContextSimplifier._simplify_pass(expression, context)
+                # debug_i += 1
+                # if debug_i > 100:
+                #     raise ValueError()
+            return new_expression
--- a/neuralpp/symbolic/expression.py
+++ b/neuralpp/symbolic/expression.py
@@ -406,6 +406,9 @@ class FunctionApplication(Expression, ABC):
                              f"but you are setting {i-1}th arguments.")
 
     def replace(self, from_expression: Expression, to_expression: Expression) -> Expression:
+        if from_expression.syntactic_eq(self):
+            return to_expression
+
         # recursively do the replacement
         new_subexpressions = [
             to_expression if e.syntactic_eq(from_expression) else e.replace(from_expression, to_expression)
--- a/neuralpp/symbolic/parameters.py
+++ b/neuralpp/symbolic/parameters.py
@@ -1,0 +1,46 @@
+"""Thread-safe global parameters. This file mimics sympy.core.parameters. """
+
+from contextlib import contextmanager
+from threading import local
+
+
+class _GlobalParameters(local):
+    """
+    Thread-local global parameters.
+
+    Explanation
+    ===========
+
+    This class generates thread-local container for SymPy's global parameters.
+    Every global parameters must be passed as keyword argument when generating
+    its instance.
+    A variable, `global_parameters` is provided as default instance for this class.
+
+    References
+    ==========
+
+    .. [1] https://docs.python.org/3/library/threading.html
+    .. [2] sympy.core.parameters._global_parameters
+
+    """
+    def __init__(self, **kwargs):
+        self.__dict__.update(kwargs)
+
+    def __setattr__(self, name, value):
+        return super().__setattr__(name, value)
+
+
+global_parameters = _GlobalParameters(sympy_evaluate=False)
+
+
+@contextmanager
+def sympy_evaluate(x):
+    old = global_parameters.sympy_evaluate
+    try:
+        global_parameters.sympy_evaluate = x
+        yield
+    except Exception as e:
+        print(f"raise: {e}")
+        raise
+    finally:
+        global_parameters.sympy_evaluate = old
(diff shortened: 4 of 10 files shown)


neuralpp/symbolic/context_simplifier.py
```diff
@@ -0,0 +1,97 @@
...
+                result = result.replace(boolean_subexpression, simplified_subexpression)
+                if result is None:
+                    raise
+
+        # replace variables
+        for variable, replacement in context.variable_replacement_dict.items():
+            result = result.replace(variable, replacement)
+        return result
+
+    def simplify(self, expression: Expression, context: Z3SolverExpression) -> Expression:
```
hsiang-wu, line 80: This is the core algorithm in this PR's description.

neuralpp/symbolic/context_simplifier.py
```diff
@@ -0,0 +1,97 @@
...
+            raise
+        for boolean_subexpression in all_boolean_subexpressions:
+            simplified_subexpression = _simplify_expression(boolean_subexpression, context)
+            if simplified_subexpression is not None:
+                result = result.replace(boolean_subexpression, simplified_subexpression)
+                if result is None:
+                    raise
+
+        # replace variables
+        for variable, replacement in context.variable_replacement_dict.items():
```
hsiang-wu, line 76: `variable_replacement_dict` is a property of `Z3SolverExpression`, where we perform the step

> Suppose there is a total order O on variables and constants where constants come before variables. Determine equivalence classes of variables and constants that are equaled by the context. Replace each variable by the minimum of its equivalent class according to O.

neuralpp/symbolic/z3_expression.py
```diff
@@ -498,7 +512,115 @@ def __and__(self, other: Any) -> Context:
...
             # Always treat `other` as a literal. Will raise if it cannot be converted to a boolean in Z3.
             new_solver = z3_add_solver_and_literal(self._solver, other.z3_object)
-            new_dict = self._dict | _extract_key_value_from_assertions([other.z3_object])
+            new_dict = self._dict | _extract_key_to_value_from_assertions([other.z3_object])
         return Z3SolverExpression.make(new_solver, new_dict)
 
     __rand__ = __and__
+
+    @cached_property
+    def variable_replacement_dict(self):
```
hsiang-wu, line 521: This property is dependent on our new class `EquivalenceClass`, see comments here for more detailed description.

neuralpp/symbolic/z3_expression.py
```diff
@@ -498,7 +512,115 @@ def __and__(self, other: Any) -> Context:
...
+            raise TypeError("Can only compare two OrderedZ3Expressions.")
+        if self.expression_level < other.expression_level:
+            return True
+        elif other.expression_level < self.expression_level:
+            return False
+        else:  # same level
+            return str(self) < str(other)
+
+
+def _extract_equivalence_classes_from_assertions(assertions: List[z3.ExprRef] | z3.AstVector) \
```
hsiang-wu, line 599: Returns all unique "equivalence classes"

neuralpp/symbolic/z3_expression.py
```diff
@@ -377,31 +383,39 @@ def number_of_arguments(self) -> int:
...
 
 
-def _extract_key_value_from_assertions_helper(assertions: List[z3.ExprRef] | z3.AstVector, dict_: Dict[str, Any]):
-    def is_value(v):
-        return z3.is_int_value(v) or z3.is_rational_value(v) or z3.is_algebraic_value(v) or z3.is_fp_value(v)
-
-    def is_key(k):
-        return z3.is_const(k) and not is_value(k)
-
+def _traverse_equalities(assertions: List[z3.ExprRef], accumulator: Callable[[z3.ExprRef, z3.ExprRef], type(None)]):
```
hsiang-wu, line 386: Rewrote this function to make it general-purpose, with an `accumulator`, so other traversing functions can use this.

neuralpp/test/quick_tests/symbolic/context_simplifier_test.py
```diff
@@ -0,0 +1,49 @@
...
+    context = context & (x == y)
+    context = context & (y > 4)
+    expr = if_then_else(y < if_then_else(x < 3, 1, 2), x * y, x + y)
+    result = si.simplify(expr, context)
+    assert result.function.value == operator.add
+    assert result.arguments[0].name == 'x'
+    assert result.arguments[1].name == 'y'
+
+
+def test_sympy_bug():
```
hsiang-wu, line 37: This is likely a bug in sympy, which causes me to refactor my code a bit

neuralpp/symbolic/context_simplifier.py
```diff
@@ -0,0 +1,97 @@
...
+            result = result.replace(variable, replacement)
+        return result
+
+    def simplify(self, expression: Expression, context: Z3SolverExpression) -> Expression:
+        with sympy_evaluate(True):
```
hsiang-wu, line 81: To work around the bug, I introduced a thread-local global parameter, `sympy_evaluate`, which by default is set to `False` (since we don't want sympy to evaluate). But in simplification, this can be set to `True`.

neuralpp/symbolic/parameters.py
```diff
@@ -0,0 +1,46 @@
+"""Thread-safe global parameters. This file mimics sympy.core.parameters. """
```
hsiang-wu, line 1: A file that introduces global parameter.

neuralpp/symbolic/constants.py
```diff
@@ -16,20 +23,18 @@ def if_then_else_function(type_: Type) -> Expression:
...
-            return bool_if_then_else_function(if_, then_, else_)
-        case int():
-            return int_if_then_else_function(if_, then_, else_)
-        case float():
-            return float_if_then_else_function(if_, then_, else_)
-        case fractions.Fraction():
-            return real_if_then_else_function(if_, then_, else_)
-        case _:
-            raise TypeError(f"Unrecognized type {type(then_)}.")
+def if_then_else(if_: Any, then_: Any, else_: Any):
```
hsiang-wu, line 26: Also redesigned `if_then_else` so it can take not only python objects but also our `Expression`s.

neuralpp/symbolic/expression.py
```diff
@@ -406,6 +406,9 @@ def set(self, i: int, new_expression: Expression) -> Expression:
                              f"but you are setting {i-1}th arguments.")
 
     def replace(self, from_expression: Expression, to_expression: Expression) -> Expression:
+        if from_expression.syntactic_eq(self):
+            return to_expression
```
hsiang-wu, line 410: This is a bug fix. Previous version cannot replace self.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -36,7 +36,6 @@ def _infer_sympy_object_type(sympy_object: sympy.Basic, type_dict: Dict[sympy.Ba
             return fractions.Fraction
         case sympy.logic.boolalg.BooleanAtom():
             return bool
-        # TODO: delete me
```
hsiang-wu: Line #39 and #40 cannot be deleted because otherwise we cannot recognize an internal class of sympy.Piecewise (`sympy.functions.elementary.piecewise.ExprCondPair`)

neuralpp/test/quick_tests/symbolic/context_simplifier_test.py
```diff
@@ -0,0 +1,49 @@
...
+    x, y = sympy.symbols('x y')
```
rodrigodesalvobraz, line 38: I don't quite see what the bug is exactly, or how to avoid it. Can you please add comments explaining what the bug is, and how you avoided it in your code? Also, can you please add comments at the point of the code where you do something to avoid this bug?

rodrigodesalvobraz (changes requested):
Really cool PR, the code is so elegant!

neuralpp/symbolic/context_simplifier.py
```diff
@@ -0,0 +1,97 @@
...
+    context implies expression iff (context => expression) is valid;
+    which means not (context => expression) is unsatisfiable;
+    which means not (not context or expression) is unsatisfiable;
+    which means context and not expression is unsatisfiable.
+    """
+    new_context = context & ~expression
+    if new_context.satisfiability_is_known:
+        return new_context.unsatisfiable
+    else:
+        return False
```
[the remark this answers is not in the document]
hsiang-wu: Yeah, that's a better name

neuralpp/test/quick_tests/symbolic/context_simplifier_test.py
```diff
@@ -36,14 +36,20 @@ def test_context_simplifier2():
...
         with sympy.evaluate(False):
             sympy.Piecewise((x*y, y < sympy.Piecewise((1, x < 3), (2, True))), (x+y, True))
 
-    with sympy.evaluate(True):
-        sympy_Cond(y < sympy.Piecewise((1, x < 3), (2, True)), x*y, x+y)
-        sympy.Piecewise((x * y, y < sympy.Piecewise((1, x < 3), (2, True))), (x + y, True))
-
+    # This means we cannot create such expressions in SymPy unless it is fixed in the library.
+    # However, for simplify() methods, we don't need to stop SymPy from evaluating, so we can set evaluate=True
+    # and work around this bug at least for simplify()'s cases.
```
hsiang-wu, line 55: Added more comments of the SymPy bug

neuralpp/symbolic/context_simplifier.py
```diff
@@ -78,7 +75,7 @@ def _simplify_pass(expression: Expression, context: Z3SolverExpression) -> Expre
         return result
 
     def simplify(self, expression: Expression, context: Z3SolverExpression) -> Expression:
-        with sympy_evaluate(True):
+        with sympy_evaluate(True):  # To work around a bug in SymPy (see context_simplifier_test.py/test_sympy_bug).
```
hsiang-wu, line 78: Here's the workaround

neuralpp/symbolic/sympy_interpreter.py
```diff
@@ -48,7 +48,7 @@ def simplify(self, expression: Expression, context: Context = TrueContext()) ->
         """
         The function calls simplify() from sympy library and wrap the result in SymPyExpression.
         """
-        with sympy_evaluate(True):
+        with sympy_evaluate(True):  # To work around a bug in SymPy (see context_simplifier_test.py/test_sympy_bug).
```
hsiang-wu, line 51: Another workaround

neuralpp/symbolic/sympy_expression.py
```diff
@@ -329,22 +335,28 @@ def native_arguments(self) -> Tuple[sympy.Basic, ...]:
...
             # see test/sympy_test.py: test_sympy_bug()
             sympy_object = sympy_function(*[sympy_argument.sympy_object for sympy_argument in sympy_arguments],
-                                          evaluate=False)
+                                          evaluate=global_parameters.sympy_evaluate)
         elif sympy_function == sympy.Piecewise:
             sympy_object = sympy_piecewise_from_if_then_else(
                 *[sympy_argument.sympy_object for sympy_argument in sympy_arguments])
         else:
-            with sympy.evaluate(False):
+            with sympy.evaluate(global_parameters.sympy_evaluate):
```
hsiang-wu, line 351: Instead of always open the context with `sympy.evaluate(False)`, we set the value to our global parameter `global_parameters.sympy_evaluate`, which if call by `simplify()`, the value is set to True to allow evaluation.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -329,22 +335,28 @@ def native_arguments(self) -> Tuple[sympy.Basic, ...]:
 
     @staticmethod
     def from_sympy_function_and_general_arguments(sympy_function: sympy.Basic, arguments: List[Expression]) -> \
-            SymPyFunctionApplication:
+            SymPyExpression:
```
hsiang-wu, line 338: Since the function can possibly evaluate the SymPy object, it does not necessarily return a function application now.

neuralpp/symbolic/parameters.py
hsiang-wu: A facility to work around the SymPy bug.

hsiang-wu (commented):
More comments on the bug and the workaround

neuralpp/symbolic/context_simplifier.py
```diff
@@ -0,0 +1,94 @@
...
+                raise Context.UnknownError()
+            if context.unsatisfiable:
+                raise ValueError(f"Context {context} is unsatisfiable.")
+
+            new_expression = ContextSimplifier._simplify_pass(expression, context)
+            # debug_i = 0
+            while not new_expression.syntactic_eq(expression):
+                expression = new_expression
+                new_expression = ContextSimplifier._simplify_pass(expression, context)
+                # debug_i += 1
```
CactusWin, line 91: Let's remove these debug statements? (Line 87, 91 - 93)

CactusWin (commented):
Looks good! I'm excited for all the bug fixes and for the context simplifier :)

neuralpp/symbolic/context_simplifier.py
hsiang-wu: Yes.

neuralpp/test/quick_tests/symbolic/context_simplifier_test.py
```diff
@@ -0,0 +1,68 @@
...
+    x = Z3Variable(z3.Int('x'))
+    y = Z3Variable(z3.Int('y'))
+    context = Z3SolverExpression()
+    context = context & (x == y)  # parenthesis is necessary
+    context = context & (y > 4)
+    result = si.simplify(sympy_x > 3, context)
+    assert result.value  # which means result is simplified to "True"
+
+
+def test_context_simplifier2():
```
rodrigodesalvobraz, line 23: I am thinking we need more unit tests since these are very simple. For example, tests involving the simplification of a bound variable to a constant or to a function application.
Another one: what happens to expression "x" if context is "y = 3 and x = y"?

rodrigodesalvobraz (changes requested):
I just realized it would be good to have a little more complex unit tests.

neuralpp/symbolic/z3_expression.py
```diff
@@ -413,25 +427,25 @@ def dict(self) -> Dict[str, Any]:
     def syntactic_eq(self, other) -> bool:
         return False  # why do we need to compare two Z3ConjunctiveClause?
 
-    def __init__(self, z3_solver: z3.Solver, value_dict: Dict[str, Any] | None = None):
+    def __init__(self, z3_solver: z3.Solver = z3.Solver(), value_dict: Dict[str, Any] | None = None):
```
horizon-blue, line 430: We aren't going to mutate `z3_solver` in `Z3SolverExpression`, right? If so, this is fine; otherwise, we should be careful about having a mutable object as the default argument [as it can lead to unexpected outcomes](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)

neuralpp/symbolic/z3_expression.py
```diff
@@ -498,7 +512,115 @@ def __and__(self, other: Any) -> Context:
...
+
+    __hash__ = frozenset.__hash__  # otherwise it's set to None by default
+
+    @cached_property
+    def minimum(self) -> Any:
+        """ Assumes len(self) > 0. """
+        min_value = None
+        for element in self:
+            min_value = element if min_value is None else min(element, min_value)
+        return min_value
```
horizon-blue, line 563: Looks like we can simply return `min(self)` here?

horizon-blue (commented):
:) LGTM other than some minor points

neuralpp/symbolic/context_simplifier.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @staticmethod
+    def _simplify_pass(expression: Expression, context: Z3SolverExpression) -> Expression:
+        result = ContextSimplifier.sympy_interpreter.simplify(expression, context)
+        assert result is not None
+
+        # replace boolean expressions
+        all_boolean_subexpressions = _collect_subset_expressions(result, lambda expr: expr.type == bool)
+        if result is None:
+            raise
```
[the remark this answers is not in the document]
hsiang-wu: > Does _collect_subset_expressions modify result?

No. I'll replace the first `assert` with this `raise`

neuralpp/symbolic/parameters.py
```diff
@@ -0,0 +1,46 @@
...
+
+@contextmanager
+def sympy_evaluate(x):
+    old = global_parameters.sympy_evaluate
+    try:
+        global_parameters.sympy_evaluate = x
+        yield
+    except Exception as e:
+        print(f"raise: {e}")
+        raise
```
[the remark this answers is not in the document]
hsiang-wu: Yes, it's redundant. I put it there when I was debugging. Initially thought context manager will not raise if I don't have this `except` clause (but that's not true).

neuralpp/symbolic/z3_expression.py
hsiang-wu: Yes. The old design here contains a bug so I rewrote this part. In fixed version I use `min()`.

rodrigodesalvobraz (approved):

PR #25, opened 2022-07-08
Symbolic Factor If-Then-Else and Variable Elimination Tests

CactusWin:
Added If-Then-Else tests for symbolic factor to make sure the functions work as expected in an if-then-else scenario.
Next PR:
- Since the if-then-else works, I'll create a model using if-then-else, and try to use symbolic factors on it

### DIFF
--- a/neuralpp/test/quick_tests/graphical_model/representation/factor/symbolic_factor_test.py
+++ b/neuralpp/test/quick_tests/graphical_model/representation/factor/symbolic_factor_test.py
@@ -1,6 +1,4 @@
-import pytest
 import sympy
-import operator
 from neuralpp.inference.graphical_model.representation.factor.symbolic_factor import SymbolicFactor
 from neuralpp.symbolic.sympy_expression import SymPyVariable
 from neuralpp.inference.graphical_model.variable.integer_variable import IntegerVariable
@@ -23,7 +21,9 @@ def test_sympy_condition():
 
     symbolic = SymbolicFactor([x, y], expression1)
     conditioned = symbolic.condition({x: 1})
-    assert conditioned.expression == SymPyVariable(y_symbol, int)
+
+    expected = SymPyVariable(y_symbol, int)
+    assert conditioned.expression == expected
 
 def test_sympy_if_then_else_condition():
     x = IntegerVariable("x", 3)
@@ -37,9 +37,11 @@ def test_sympy_if_then_else_condition():
 
     symbolic = SymbolicFactor([x, y], expression1)
     conditioned1 = symbolic.condition({x: 1})
-    assert conditioned1.expression.arguments[0].syntactic_eq(y_sympy < 1)
-    assert conditioned1.expression.arguments[1].value == 2
-    assert conditioned1.expression.arguments[2].value == 3
+
+    expected1 = if_then_else(y_sympy < 1, 2, 3)
+    assert conditioned1.expression.arguments[0] == expected1.arguments[0]
+    assert conditioned1.expression.arguments[1].value == expected1.arguments[1]
+    assert conditioned1.expression.arguments[2].value == expected1.arguments[2]
 
     conditioned1 = symbolic.condition({x: 1, y:2})
     assert conditioned1.expression == 3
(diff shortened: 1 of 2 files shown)


neuralpp/test/quick_tests/graphical_model/representation/factor/symbolic_factor_test.py
```diff
@@ -68,6 +138,61 @@ def test_normalize():
...
+    query = z
+
+    ve_result = VariableElimination().run(query, model)
+    brute_result = BruteForce().run(query, model)
+
+    difference = ve_result.expression - brute_result.expression
+    normalizer = Normalizer()
+    context = Z3SolverExpression()
+
+    assert normalizer.normalize(difference, context) == 0
```
CactusWin, line 198: Fixed this assert based on discussion with Rodrigo

neuralpp/test/quick_tests/graphical_model/representation/factor/symbolic_factor_test.py
```diff
@@ -19,6 +23,28 @@ def test_sympy_condition():
...
+
+    x_symbol, y_symbol = sympy.symbols("x y")
+    x_sympy = SymPyVariable(x_symbol, int)
+    y_sympy = SymPyVariable(y_symbol, int)
+
+    expression1 = if_then_else(x_sympy > y_sympy, 2, 3)
+
+    symbolic = SymbolicFactor([x, y], expression1)
+    conditioned1 = symbolic.condition({x: 1})
+    assert conditioned1.expression.arguments[0].syntactic_eq(y_sympy < 1)
```
rodrigodesalvobraz, line 40: This set of asserts would be simpler and more readable if you wrote
```
expected = if_then_else(y_sympy < 1, 2, 3)
assert conditioned1.expression.syntactic_eq(expected)
```

neuralpp/test/quick_tests/graphical_model/representation/factor/symbolic_factor_test.py
```diff
@@ -42,6 +68,30 @@ def test_mul_by_non_identity():
...
+    x_sympy = SymPyVariable(x_symbol, int)
+    y_sympy = SymPyVariable(y_symbol, int)
+    z_sympy = SymPyVariable(z_symbol, int)
+
+    expression1 = if_then_else(x_sympy > y_sympy, 2, 3)
+    expression2 = if_then_else(x_sympy == z_sympy, 1, 5)
+
+    symbolic1 = SymbolicFactor([x, y], expression1)
+    symbolic2 = SymbolicFactor([x, z], expression2)
+    symbolic3 = symbolic1.mul_by_non_identity(symbolic2)
```
rodrigodesalvobraz, line 89: I should have named `mul_by_non_identity` as a private method `_mul_by_non_identity`, because it is not supposed to be used externally. The operator `*` (`__mul__`) takes care of the identity case and calls `mul_by_non_identity` to deal with other cases, and that is the only place where this method should be called. If called elsewhere, like here, it may fail because it doesn't deal with the identity case necessarily.
In this case here you can simply write `symbolic3 = symbolic1 * symbolic2`.

neuralpp/test/quick_tests/graphical_model/representation/factor/symbolic_factor_test.py
```diff
@@ -54,8 +104,26 @@ def test_sum_out_variable():
...
+    y = IntegerVariable("y", 2)
+
+    x_symbol, y_symbol = sympy.symbols("x y")
+    x_sympy = SymPyVariable(x_symbol, int)
+    y_sympy = SymPyVariable(y_symbol, int)
+
+    expression1 = if_then_else(x_sympy > y_sympy, 2, 3)
+
+    symbolic = SymbolicFactor([x, y], expression1)
+    sum_out_x = symbolic.sum_out_variable(x)
```
rodrigodesalvobraz, line 123: Suggestion: you can use `^` for `sum_out_variable`: `symbolic ^ x`. It also works if the second argument is an `Iterable` of variables.

rodrigodesalvobraz (changes requested):

neuralpp/test/quick_tests/graphical_model/representation/factor/symbolic_factor_test.py
```diff
@@ -35,9 +37,11 @@ def test_sympy_if_then_else_condition():
 
     symbolic = SymbolicFactor([x, y], expression1)
     conditioned1 = symbolic.condition({x: 1})
-    assert conditioned1.expression.arguments[0].syntactic_eq(y_sympy < 1)
-    assert conditioned1.expression.arguments[1].value == 2
-    assert conditioned1.expression.arguments[2].value == 3
+
+    expected1 = if_then_else(y_sympy < 1, 2, 3)
+    assert conditioned1.expression.arguments[0] == expected1.arguments[0]
```
rodrigodesalvobraz, line 42: Why not simply `conditioned1.expression.syntactic_eq(expected1)`?

neuralpp/test/quick_tests/graphical_model/representation/factor/symbolic_factor_test.py
```diff
@@ -21,7 +21,9 @@ def test_sympy_condition():
 
     symbolic = SymbolicFactor([x, y], expression1)
     conditioned = symbolic.condition({x: 1})
-    assert conditioned.expression == SymPyVariable(y_symbol, int)
+
+    expected = SymPyVariable(y_symbol, int)
+    assert conditioned.expression == expected
```
rodrigodesalvobraz, line 26: This is probably not really checking if the expressions are equal, it is just making a new `Expression` that is an equality. We need to use `syntactic_eq` for all assertions of `Expression` comparisons.

rodrigodesalvobraz (changes requested):

rodrigodesalvobraz (approved):
:-)

PR #26, opened 2022-07-08
normalization

hsiang-wu:
- Wrote a utility method that returns the first formula occurring in a given expression in a breadth-first search.
- Defined a class Normalizer with a method `normalize(e: Expression, context: Z3SolverExpression) -> Expression` that does the following:
  - Simplify the expression `e`.
  - Search for the first atom in the expression `e`.
  - If there is no atom in the expression `e`, return `e`. Otherwise:
  - Compute `context1` obtained by conjoining the atom with `context`.
  - Compute `context2` obtained by conjoining the negation of the atom with `context`.
  - Obtain `e1` by replacing atom by `True` in `e`.
  - Obtain `e2` by replacing atom by `False` in `e`.
  - Obtain `n1` by normalizing `e1` with context `c1`.
  - Obtain `n2` by normalizing `e2` with context `c2`.
  - Return expression `if atom then n1 else n2`.

### DIFF
--- a/neuralpp/symbolic/expression.py
+++ b/neuralpp/symbolic/expression.py
@@ -117,7 +117,32 @@ class Expression(ABC):
 
     @abstractmethod
     def syntactic_eq(self, other) -> bool:
+        """
+        Returns if self and other are syntactically equal, i.e.,
+        that they are of subclass of Expression and their internal representation are equal.
+        This method usually depends on subclass-specific library calls,
+        e.g., Z3Expression.syntactic_eq() would leverage z3.eq().
+        This method should be considered as a cheap way to check syntactic equality of two symbolic expressions.
+        """
         pass
+
+    def structure_eq(self, other) -> bool:
+        """
+        Returns if self and other are "structurally" equivalent, i.e., that they have the same Expression interfaces.
+        E.g, a Z3Expression of "a + b" does not syntactic_eq() a SymPyExpression of "a + b", but a call of
+        structure_eq() on the two should return True.
+        This method is general and more expensive than syntactic_eq()
+        """
+        match self, other:
+            case AtomicExpression(base_type=self_base_type, atom=self_atom, type=self_type),\
+                    AtomicExpression(base_type=other_base_type, atom=other_atom, type=other_type):
+                return self_base_type == other_base_type and self_type == other_type and self_atom == other_atom
+            case FunctionApplication(subexpressions=self_subexpressions, type=self_type), \
+                    FunctionApplication(subexpressions=other_subexpressions, type=other_type):
+                return all(lhs.structure_eq(rhs) for lhs, rhs in zip(self_subexpressions, other_subexpressions)) and \
+                       self_type == other_type
+            case _:
+                return False
 
     @classmethod
     @abstractmethod
@@ -363,7 +388,7 @@ class Constant(AtomicExpression, ABC):
         return self.atom
 
     def __str__(self) -> str:
-        return f"{self.value}: {self.type}"
+        return f"{self.value}"
 
 
 class FunctionApplication(Expression, ABC):
@@ -397,13 +422,13 @@ class FunctionApplication(Expression, ABC):
             return self.new_function_application(new_expression, self.arguments)
 
         # evaluate len after i != 0, if i == 0 we can be lazy
-        if i-1 < self.number_of_arguments:
+        if i - 1 < self.number_of_arguments:
             arguments = self.arguments
-            arguments[i-1] = new_expression
+            arguments[i - 1] = new_expression
             return self.new_function_application(self.function, arguments)
         else:
             raise IndexError(f"Out of scope. Function application only has {self.number_of_arguments} arguments "
-                             f"but you are setting {i-1}th arguments.")
+                             f"but you are setting {i - 1}th arguments.")
 
     def replace(self, from_expression: Expression, to_expression: Expression) -> Expression:
         if from_expression.syntactic_eq(self):
--- a/neuralpp/symbolic/normalizer.py
+++ b/neuralpp/symbolic/normalizer.py
@@ -1,0 +1,68 @@
+from typing import Optional, Deque
+from collections import deque
+
+from .expression import Expression, Constant, FunctionApplication
+from .z3_expression import Z3SolverExpression
+from .constants import basic_true, basic_false, if_then_else
+from .context_simplifier import ContextSimplifier
+from .parameters import sympy_evaluate
+import neuralpp.symbolic.functions as functions
+
+
+def is_boolean_atom(expression: Expression) -> bool:
+    return expression.type == bool and not isinstance(expression, Constant)  # should ignore `True` and `False`
+
+
+def _bfs_first_atom_of_expression(expression: Expression, queue: Deque) -> Optional[Expression]:
+    queue.append(expression)
+    while queue:
+        current_expression = queue.popleft()
+        if is_boolean_atom(current_expression):
+            return current_expression
+        match current_expression:
+            case FunctionApplication(function=Constant(value=functions.conditional), arguments=arguments):
+                queue.extend(arguments[1:])  # do not include the if clause, it's meaningless
+            case FunctionApplication(arguments=arguments):
+                queue.extend(arguments)
+            case _:
+                pass  # atomic expression has no subexpressions
+    return None
+
+
+def first_atom_of_expression(expression: Expression) -> Optional[Expression]:
+    """ returns the first atom occurring in a given expression in a breadth-first search. """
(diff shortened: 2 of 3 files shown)


neuralpp/symbolic/expression.py
```diff
@@ -117,8 +117,33 @@ def contains(self, target: Expression) -> bool:
...
+        """
         pass
 
+    def structure_eq(self, other) -> bool:
+        """
+        Returns if self and other are "structurally" equivalent, i.e., that they have the same Expression interfaces.
+        E.g, a Z3Expression of "a + b" does not syntactic_eq() a SymPyExpression of "a + b", but a call of
+        structure_eq() on the two should return True.
+        This method is general and more expensive than syntactic_eq()
+        """
```
hsiang-wu, line 135: Introduced a new `structure_eq` method to help writing tests. Don't know how the name sounds, I'm also thinking of `interface_eq`, suggestions?

neuralpp/symbolic/expression.py
```diff
@@ -363,7 +388,7 @@ def value(self) -> Any:
         return self.atom
 
     def __str__(self) -> str:
-        return f"{self.value}: {self.type}"
+        return f"{self.value}"
```
hsiang-wu, line 391: Got rid of type information in printing: it looks better when debugging

neuralpp/symbolic/normalizer.py
```diff
@@ -0,0 +1,68 @@
...
+def first_atom_of_expression(expression: Expression) -> Optional[Expression]:
+    """ returns the first atom occurring in a given expression in a breadth-first search. """
+    return _bfs_first_atom_of_expression(expression, deque())
+
+
+class Normalizer:
+    simplifier = ContextSimplifier()
+
+    @staticmethod
+    def _normalize(expression: Expression, context: Z3SolverExpression) -> Expression:
```
hsiang-wu, line 41: A faithful implementation of the specification

neuralpp/symbolic/normalizer.py
```diff
@@ -0,0 +1,68 @@
...
+
+def _bfs_first_atom_of_expression(expression: Expression, queue: Deque) -> Optional[Expression]:
+    queue.append(expression)
+    while queue:
+        current_expression = queue.popleft()
+        if is_boolean_atom(current_expression):
+            return current_expression
+        match current_expression:
+            case FunctionApplication(function=Constant(value=functions.conditional), arguments=arguments):
+                queue.extend(arguments[1:])  # do not include the if clause, it's meaningless
```
hsiang-wu, line 24: ignore the if clause to avoid re-creating the same conditional expression

neuralpp/symbolic/normalizer.py
```diff
@@ -0,0 +1,68 @@
...
+from .expression import Expression, Constant, FunctionApplication
+from .z3_expression import Z3SolverExpression
+from .constants import basic_true, basic_false, if_then_else
+from .context_simplifier import ContextSimplifier
+from .parameters import sympy_evaluate
+import neuralpp.symbolic.functions as functions
+
+
+def is_boolean_atom(expression: Expression) -> bool:
+    return expression.type == bool and not isinstance(expression, Constant)  # should ignore `True` and `False`
```
hsiang-wu, line 13: We should not split on `True`/`False` as it is meaningless and would cause the algorithm to not terminate. E.g., split on `True` would be `if True then True else False`...

neuralpp/symbolic/normalizer.py
```diff
@@ -0,0 +1,68 @@
...
+    simplifier = ContextSimplifier()
+
+    @staticmethod
+    def _normalize(expression: Expression, context: Z3SolverExpression) -> Expression:
+        simplified_expression = Normalizer.simplifier.simplify(expression, context)
+        first_atom = first_atom_of_expression(simplified_expression)
+        if first_atom is None:
+            return simplified_expression
+
+        print(f"first_atom {first_atom}")
```
CactusWin, line 47: Was this print statement for debugging?

neuralpp/symbolic/expression.py
horizon-blue: `structure_eq` sounds quite reasonable to me :)

neuralpp/symbolic/normalizer.py
horizon-blue: :) Thanks for the clarification. It'd be helpful to include your explanation here as comments, so future readers wouldn't be confused

neuralpp/symbolic/normalizer.py
```diff
@@ -0,0 +1,67 @@
...
+                queue.extend(arguments[1:])  # do not include the if clause, it's meaningless
+            case FunctionApplication(arguments=arguments):
+                queue.extend(arguments)
+            case _:
+                pass  # atomic expression has no subexpressions
+    return None
+
+
+def first_atom_of_expression(expression: Expression) -> Optional[Expression]:
+    """ returns the first atom occurring in a given expression in a breadth-first search. """
```
horizon-blue, line 33: Probably worth mentioning that we're looking for the first atom *that's not a constant*

neuralpp/test/quick_tests/symbolic/normalizer_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+    context = context & (a == b)
+    result2 = normalizer.normalize(expr, context)
+    assert result2.value == 3
+
+
+def test_normalizer2():
+    """
+    f: bool -> bool -> int -> int
+    f(a > b, b > c, 5)
```
horizon-blue, line 29: typo: `a < b`?

neuralpp/test/quick_tests/symbolic/normalizer_test.py
```diff
@@ -0,0 +1,63 @@
...
+    a = BasicVariable('a', int)
+    b = BasicVariable('b', int)
+    c = BasicVariable('c', bool)
+
+    expr = f(c, f((a > b) | c, 3))
+    context = Z3SolverExpression()
+    context = context & (f(True, 66) == 45) & (f(True, 3) == 66) & (f(False, 3) == 3)
+    assert isinstance(context, Z3SolverExpression)
+    result = normalizer.normalize(expr, context)
+    assert result.structure_eq(if_then_else(c, 45, if_then_else(a > b, f(False, 66), 3)))
```
horizon-blue, line 63: wow, thinking through this one is hard. It's really amazing that we are able to get this correct :D

horizon-blue (commented):
Super neat! I just have a few minor points as inline comments. Other than those this PR looks great to me

neuralpp/symbolic/normalizer.py
hsiang-wu: I think the current definition of `atom` is not very clear as it can be "non-constant expression" (as used in `is_boolean_atom`), or "non-constant boolean expression" (as used in `first_atom_of_expression`), or "boolean expression" (as proposed by you here, if we change the function name to `first_non_constant_atom_of_expression`). 

Also, since "atom" is not a very straight-forward name, we've switched to "formula" in the project plan. I think we should do the same here. "Formula" is a better understood name, which should mean "boolean expression".

neuralpp/test/quick_tests/symbolic/normalizer_test.py
hsiang-wu: I'm adding some comments to walk readers through this example.

neuralpp/symbolic/normalizer.py
```diff
@@ -40,26 +40,29 @@ class Normalizer:
     @staticmethod
     def _normalize(expression: Expression, context: Z3SolverExpression) -> Expression:
         simplified_expression = Normalizer.simplifier.simplify(expression, context)
-        first_atom = first_atom_of_expression(simplified_expression)
-        if first_atom is None:
+        # We look for non-constant formula, because we should not split on constants (i.e., True/False)
```
hsiang-wu, line 43: Moved a comment here since it makes more sense

neuralpp/symbolic/normalizer.py
```diff
@@ -9,29 +9,29 @@
 import neuralpp.symbolic.functions as functions
 
 
-def is_boolean_atom(expression: Expression) -> bool:
+def is_non_constant_formula(expression: Expression) -> bool:
```
hsiang-wu, line 12: now using "formula" for "boolean expressions"

neuralpp/test/quick_tests/symbolic/normalizer_test.py
horizon-blue: Thanks for doing this!

horizon-blue (approved):
Looks great to me with the updates :).

PR #27, opened 2022-07-12
quantifier expressions

hsiang-wu:
- Define a new interface QuantifiedExpression subclassing Expression. It should provide:
  - a property `operation` providing the function on which the quantifier is based (for example, or for existential quantification, and for universal quantification, + for summation, * for product).
  - a property `index` for the variable over which this quantifier iterates.
  - a property `constraint` for the formula expression determining the values of the index over which the quantier iterates.
  - a property `body` for the expression being quantified.
  - Example: `sum_{x : x != 3 and x > 1} x^1 + 2`
- Define an implementation of `BasicQuantifier`.
- Define a class `Summation` subclassing BasicQuantifier for operator `+`.
- Write unit tests.

### DIFF
--- a/neuralpp/symbolic/basic_expression.py
+++ b/neuralpp/symbolic/basic_expression.py
@@ -3,9 +3,10 @@ from __future__ import annotations
 import operator
 import builtins
 from neuralpp.symbolic.expression import Expression, FunctionApplication, Variable, Constant, AtomicExpression, \
-    VariableNotTypedError, ExpressionType, get_arithmetic_function_type_from_argument_types, Context
+    VariableNotTypedError, ExpressionType, get_arithmetic_function_type_from_argument_types, Context, \
+    QuantifierExpression
 from abc import ABC
-from typing import Any, List, Optional, Callable, Dict
+from typing import Any, List, Optional, Callable, Dict, get_args
 import neuralpp.symbolic.functions as functions
 
 
@@ -72,6 +73,11 @@ class BasicExpression(Expression, ABC):
     @classmethod
     def new_function_application(cls, function: Expression, arguments: List[Expression]) -> BasicFunctionApplication:
         return BasicFunctionApplication(function, arguments)
+
+    @classmethod
+    def new_quantifier_expression(cls, operation: Constant, index: Variable, constrain: Expression, body: Expression,
+                                  ) -> QuantifierExpression:
+        return BasicQuantifierExpression(operation, index, constrain, body)
 
 
 class BasicAtomicExpression(BasicExpression, AtomicExpression, ABC):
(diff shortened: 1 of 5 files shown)


neuralpp/symbolic/basic_expression.py
```diff
@@ -167,8 +173,53 @@ def number_of_arguments(self) -> int:
 
     def syntactic_eq(self, other) -> bool:
         match other:
-            case BasicFunctionApplication(function=function, arguments=arguments, type=other_type):
-                return all(lhs.syntactic_eq(rhs) for lhs, rhs in zip(self.subexpressions, [function] + arguments)) and \
-                       self.type == other_type
+            case BasicFunctionApplication(subexpressions=other_subexpressions):
```
hsiang-wu, line 176: We don't need to compare type since the equivalence of `function` implies the equivalence of type of the function application, if the argument numbers are equal.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -206,6 +206,11 @@ def new_function_application(cls, function: Expression, arguments: List[Expressi
             case _:
                 raise ValueError("Unknown case.")
 
+    @classmethod
+    def new_quantifier_expression(cls, operation: Constant, index: Variable, constrain: Expression, body: Expression,
+                                  ) -> QuantifierExpression:
+        raise NotImplementedError()
```
hsiang-wu, line 212: Placeholder to appease ABC

neuralpp/symbolic/expression.py
```diff
@@ -471,6 +479,51 @@ def dict(self) -> Dict[str, Any]:
...
+        subexpressions[i] = new_expression
+        return self.new_quantifier_expression(*subexpressions)
+
+    def replace(self, from_expression: Expression, to_expression: Expression) -> Expression:
+        if from_expression.syntactic_eq(self):
+            return to_expression
+
+        # recursively do the replacement
+        new_subexpressions = [
+            to_expression if e.syntactic_eq(from_expression) else e.replace(from_expression, to_expression)
```
horizon-blue, line 521: We might not even need to check `e.syntactic_eq` here because `e.replace` cover that case already. (ditto to the definition of `FunctionApplication.replace`)

neuralpp/test/quick_tests/symbolic/expression_test.py
```diff
@@ -137,6 +137,21 @@ def test_basic_function_application():
...
+    assert sum_.subexpressions[1].syntactic_eq(i)
+    assert sum_.subexpressions[2].syntactic_eq(i_range)
+    assert sum_.subexpressions[3].syntactic_eq(i)
+    assert len(sum_.subexpressions) == 4
+    a = BasicVariable('a', int)
+    new_sum = sum_.replace(i, a)
+    assert sum_.syntactic_eq(BasicSummation(int, i, (0 < i) & (i < 10), i))
+    assert new_sum.syntactic_eq(BasicSummation(int, a, (0 < a) & (a < 10), a))
+
+
```
horizon-blue, line 154: Can we have a test here for the `set` method as well? I'd also love to see an example of nested `QuantifierExpression`, which is a very common use case.

horizon-blue (commented):
Looks pretty good to me :). I just have a few minor comments

neuralpp/test/quick_tests/symbolic/expression_test.py
hsiang-wu: Yes!

CactusWin (commented):
Looks good to me!

CactusWin (approved):

horizon-blue (approved):
Looks great! Thanks for the quick update :). Looking forward to see more works to be done on quantifier expressions

PR #29, opened 2022-07-14
normalization of quantifier expression

hsiang-wu:
Write an algorithm that, given `Quant_{x: F} E` and context `C`, rewrites it as follows:
- `N <- normalize(E, C and F)`
- if `N` is not an if then else expression, return `Quant_{x: F} N`. Otherwise:
- `N` is of the form `if Cond then N1 else N2`
- If `Cond` contains index `x`
   - Return `(Quant_{x: F and Cond} N1) + (Quant_{x: F and not Cond} N2)`
- else
   - Return `if Cond then Quant_{x: F} N1 else Quant_{x: F} N2`

### DIFF
--- a/neuralpp/symbolic/basic_expression.py
+++ b/neuralpp/symbolic/basic_expression.py
@@ -4,7 +4,7 @@ import operator
 import builtins
 from neuralpp.symbolic.expression import Expression, FunctionApplication, Variable, Constant, AtomicExpression, \
     VariableNotTypedError, ExpressionType, get_arithmetic_function_type_from_argument_types, Context, \
-    QuantifierExpression
+    QuantifierExpression, AbelianOperation
 from abc import ABC
 from typing import Any, List, Optional, Callable, Dict, get_args
 import neuralpp.symbolic.functions as functions
@@ -75,9 +75,13 @@ class BasicExpression(Expression, ABC):
         return BasicFunctionApplication(function, arguments)
 
     @classmethod
-    def new_quantifier_expression(cls, operation: Constant, index: Variable, constrain: Expression, body: Expression,
-                                  ) -> QuantifierExpression:
-        return BasicQuantifierExpression(operation, index, constrain, body)
+    def new_quantifier_expression(cls,
+                                  operation: AbelianOperation, index: Variable, constrain: Context, body: Expression,
+                                  ) -> Expression:
+        if constrain.satisfiability_is_known and constrain.unsatisfiable:
+            return operation.identity
+        else:
+            return BasicQuantifierExpression(operation, index, constrain, body)
 
 
 class BasicAtomicExpression(BasicExpression, AtomicExpression, ABC):
@@ -181,7 +185,7 @@ class BasicFunctionApplication(BasicExpression, FunctionApplication):
 
 
 class BasicQuantifierExpression(QuantifierExpression, BasicExpression):
-    def __init__(self, operation: Constant, index: Variable, constrain: Expression, body: Expression):
+    def __init__(self, operation: AbelianOperation, index: Variable, constrain: Context, body: Expression):
         self._operation = operation
         self._index = index
         self._constrain = constrain
@@ -215,11 +219,25 @@ class BasicQuantifierExpression(QuantifierExpression, BasicExpression):
                 return False
 
 
+class BasicAbelianOperation(BasicConstant, AbelianOperation):
+    @property
+    def identity(self) -> Any:
+        return self._identity
+
+    def __init__(self, operation: Callable, element_type: type, identity: Any):
+        # technically, the type of identity should be element_type, but there seems no way to declare that in Python?
+        self._identity = identity
+        super().__init__(operation, Callable[[element_type, element_type], element_type])
+
+
+def basic_add_operation(type_) -> BasicAbelianOperation:
+    return BasicAbelianOperation(operator.add, type_, 0)
+
+
 class BasicSummation(BasicQuantifierExpression):
-    def __init__(self, type_: type, index: Variable, constrain: Expression, body: Expression):
+    def __init__(self, type_: type, index: Variable, constrain: Context, body: Expression):
         """
         Expect type_ to be the argument type/return type of the summation.
         E.g., the type_ of Summation{i in [0,100]}(i) should be `int`.
         """
-        summation_operation = BasicConstant(operator.add, Callable[[type_, type_], type_])
-        super().__init__(summation_operation, index, constrain, body)
+        super().__init__(basic_add_operation(type_), index, constrain, body)
--- a/neuralpp/symbolic/basic_expression.py
+++ b/neuralpp/symbolic/basic_expression.py
@@ -76,12 +76,12 @@ class BasicExpression(Expression, ABC):
 
     @classmethod
     def new_quantifier_expression(cls,
-                                  operation: AbelianOperation, index: Variable, constrain: Context, body: Expression,
+                                  operation: AbelianOperation, index: Variable, constraint: Context, body: Expression,
                                   ) -> Expression:
-        if constrain.satisfiability_is_known and constrain.unsatisfiable:
+        if constraint.satisfiability_is_known and constraint.unsatisfiable:
             return operation.identity
         else:
-            return BasicQuantifierExpression(operation, index, constrain, body)
+            return BasicQuantifierExpression(operation, index, constraint, body)
 
 
 class BasicAtomicExpression(BasicExpression, AtomicExpression, ABC):
@@ -185,10 +185,10 @@ class BasicFunctionApplication(BasicExpression, FunctionApplication):
 
 
 class BasicQuantifierExpression(QuantifierExpression, BasicExpression):
-    def __init__(self, operation: AbelianOperation, index: Variable, constrain: Context, body: Expression):
+    def __init__(self, operation: AbelianOperation, index: Variable, constraint: Context, body: Expression):
         self._operation = operation
         self._index = index
-        self._constrain = constrain
+        self._constraint = constraint
         self._body = body
         argument_types, return_type = get_args(operation.type)
         if len(argument_types) != 2 or not (argument_types[0] == argument_types [1] == return_type):
@@ -204,8 +204,8 @@ class BasicQuantifierExpression(QuantifierExpression, BasicExpression):
         return self._index
 
     @property
-    def constrain(self) -> Expression:
-        return self._constrain
+    def constraint(self) -> Expression:
+        return self._constraint
 
     @property
     def body(self) -> Expression:
@@ -221,23 +221,24 @@ class BasicQuantifierExpression(QuantifierExpression, BasicExpression):
 
 class BasicAbelianOperation(BasicConstant, AbelianOperation):
     @property
-    def identity(self) -> Any:
+    def identity(self) -> Expression:
         return self._identity
 
-    def __init__(self, operation: Callable, element_type: type, identity: Any):
+    def __init__(self, operation: Callable, identity: Expression):
         # technically, the type of identity should be element_type, but there seems no way to declare that in Python?
         self._identity = identity
+        element_type = identity.type
         super().__init__(operation, Callable[[element_type, element_type], element_type])
 
 
 def basic_add_operation(type_) -> BasicAbelianOperation:
-    return BasicAbelianOperation(operator.add, type_, 0)
+    return BasicAbelianOperation(operator.add, BasicConstant(0, type_))
 
 
 class BasicSummation(BasicQuantifierExpression):
-    def __init__(self, type_: type, index: Variable, constrain: Context, body: Expression):
+    def __init__(self, type_: type, index: Variable, constraint: Context, body: Expression):
         """
         Expect type_ to be the argument type/return type of the summation.
         E.g., the type_ of Summation{i in [0,100]}(i) should be `int`.
         """
-        super().__init__(basic_add_operation(type_), index, constrain, body)
+        super().__init__(basic_add_operation(type_), index, constraint, body)
--- a/neuralpp/symbolic/basic_expression.py
+++ b/neuralpp/symbolic/basic_expression.py
@@ -99,7 +99,7 @@ class BasicAtomicExpression(BasicExpression, AtomicExpression, ABC):
     def atom(self) -> str:
         return self._atom
 
-    def syntactic_eq(self, other) -> bool:
+    def internal_object_eq(self, other) -> bool:
         match other:
             case BasicAtomicExpression(base_type=other_base_type, atom=other_atom, type=other_type):
                 return other_base_type == self.base_type and self.type == other_type and self.atom == other_atom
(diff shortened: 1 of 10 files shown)


neuralpp/symbolic/expression.py
```diff
@@ -106,13 +106,33 @@ def contains(self, target: Expression) -> bool:
         f(x,f(a,b)).contains(a) == True
         a.contains(a) == True
         """
-        if self == target:
+        if self.syntactic_eq(target):
```
hsiang-wu, line 109: it's a previous bug

neuralpp/symbolic/expression.py
```diff
@@ -106,13 +106,33 @@ def contains(self, target: Expression) -> bool:
...
             return True
-
         for sub_expr in self.subexpressions:
             if sub_expr.contains(target):
                 return True
+        return False
 
+    def structurally_contains(self, target: Expression) -> bool:
```
hsiang-wu, line 116: introduced this function because i need this "contains" in normalization of quantifier expression

neuralpp/symbolic/normalizer.py
```diff
@@ -38,7 +39,12 @@ class Normalizer:
     simplifier = ContextSimplifier()
 
     @staticmethod
-    def _normalize(expression: Expression, context: Z3SolverExpression) -> Expression:
+    def _basic_normalize(expression: Expression, context: Z3SolverExpression) -> Expression:
```
hsiang-wu, line 42: renamed this `_normalize` to `_basic_normalize`, since we now have a `_quantifier_expression_normalize`

neuralpp/symbolic/basic_expression.py
```diff
@@ -181,7 +185,7 @@ def syntactic_eq(self, other) -> bool:
 
 
 class BasicQuantifierExpression(QuantifierExpression, BasicExpression):
-    def __init__(self, operation: Constant, index: Variable, constrain: Expression, body: Expression):
+    def __init__(self, operation: AbelianOperation, index: Variable, constrain: Context, body: Expression):
```
hsiang-wu, line 188: Changed the types `operation` and `constrain` to be more precise

neuralpp/test/quick_tests/symbolic/expression_test.py
```diff
@@ -140,25 +141,27 @@ def test_basic_function_application():
...
     from neuralpp.symbolic.constants import int_add, int_multiply
     i = BasicVariable('i', int)
-    i_range = (0 < i) & (i < 10)
+    context = Z3SolverExpression()
+    i_range = context & (0 < i) & (i < 10)
     sum_ = BasicSummation(int, i, i_range, i)
     assert sum_.subexpressions[0].syntactic_eq(int_add)
     assert sum_.subexpressions[1].syntactic_eq(i)
-    assert sum_.subexpressions[2].syntactic_eq(i_range)
+    assert sum_.subexpressions[2].structure_eq(i_range)  # cannot check syntactic_eq() of Z3SolverExpression.
```
hsiang-wu, line 149: changed a bunch of `syntactic_eq` to `structure_eq` because `syntactic_eq()` does not work for `Z3SolverExpression` (for Z3's `syntactic_eq()`, we compare the `z3_object`; `Z3SolverExpression` does not contain a `z3_object` to begin with, so all its `syntactic_eq` calls return `False`)

neuralpp/test/quick_tests/symbolic/normalizer_test.py
```diff
@@ -78,3 +79,50 @@ def test_normalizer3():
...
+
+    context = empty_context & (i < 5)
+    with pytest.raises(ValueError):
+        normalizer.normalize(sum_, context)
+
+    j = BasicVariable('j', int)
+    sum_ = BasicSummation(int, i, empty_context & (j < i) & (i < 10), i + j)
+    context = empty_context & (j == 5)
+    assert normalizer.normalize(sum_, context).structure_eq(
+        BasicSummation(int, i, empty_context & (j < i) & (i < 10), 5 + i))  # SymPy switched 5 and i
```
horizon-blue, line 103: 🤔 Actually I was just thinking if it would make sense to replace `j` in the constrain as well. Do you have any thoughts on this?

horizon-blue (commented):
Great work :). I just have a few comments & questions

neuralpp/symbolic/basic_expression.py
```diff
@@ -215,11 +219,25 @@ def syntactic_eq(self, other) -> bool:
                 return False
 
 
+class BasicAbelianOperation(BasicConstant, AbelianOperation):
+    @property
+    def identity(self) -> Any:
+        return self._identity
+
+    def __init__(self, operation: Callable, element_type: type, identity: Any):
+        # technically, the type of identity should be element_type, but there seems no way to declare that in Python?
```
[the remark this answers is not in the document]
hsiang-wu: Does it support a use case like `def __init__(self, operation: Callable, T: type, identity: T)`? I didn't find it in the document.

The closest solution I can think of is `def __init__(self, operation: Callable, element_type: type(T), identity: T)`, which is not quite right.

neuralpp/symbolic/expression.py
```diff
@@ -164,7 +184,7 @@ def new_variable(cls, name: str, type_: ExpressionType) -> Variable:
 
     @classmethod
     @abstractmethod
-    def new_function_application(cls, function: Expression, arguments: List[Expression]) -> FunctionApplication:
+    def new_function_application(cls, function: Expression, arguments: List[Expression]) -> Expression:
```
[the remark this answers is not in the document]
hsiang-wu: That should be the correct return type, since in some past PR I changed `SymPyExpression.new_function_application()` to necessarily return `FunctionApplication`.

But I'll also need to change the return type of `new_quantifier_expression()`.

neuralpp/test/quick_tests/symbolic/normalizer_test.py
hsiang-wu: I think both are correct. If we want to have the "replacing according to the context" behavior, what we essentially need is "simplify constrain according to context", which is more general and not only deals with equalities. 

But since `constrain` is a context (`Z3SolverExpression`), simplification of `constrain` is not very efficient (first convert to `SymPyExpression`, then we need to manually convert it back to `Z3SolverExpression`, which is not implemented yet). 

I would argue since this feature (i) does not affect correctness, (ii) requires some non-trivial work to implement, and (iii) would not be very efficient, we should just be lazy here and stick to the current behavior.

neuralpp/symbolic/expression.py
```diff
@@ -476,10 +495,29 @@ def dict(self) -> Dict[str, Any]:
         pass
 
 
+class AbelianOperation(Constant, ABC):
+    """
+    An Abelian operation is a commutative, associative binary operation with an identity element.
```
CactusWin, line 500: I had to search up what Abelian meant haha (o´∀`o) Maybe we can add a link to the wiki page for people who never heard of the term before (https://en.wikipedia.org/wiki/Abelian_group). :)

neuralpp/symbolic/expression.py
```diff
@@ -476,10 +495,29 @@ def dict(self) -> Dict[str, Any]:
...
+    def element_type(self) -> ExpressionType:
+        return self.identity.type
+
+
 class QuantifierExpression(Expression, ABC):
     @property
     @abstractmethod
-    def operation(self) -> Constant:
+    def operation(self) -> AbelianOperation:
         """ operation is a Constant wrapping a function.
```
CactusWin, line 521: Very small nit: I think for the other comments, you have """ on a separate line
E.g.:
```
"""
Comment
""""
```

neuralpp/symbolic/z3_expression.py
```diff
@@ -437,6 +438,22 @@ def dict(self) -> Dict[str, Any]:
...
+    def replace(self, from_expression: Expression, to_expression: Expression) -> Z3SolverExpression:
+        """
+        If we do not override this replace(), the default replace() will causing the return value to be
+        a Z3FunctionApplication, where the result is no longer a Context.
+        """
+        from_expression = Z3Expression.convert(from_expression)
+        to_expression = Z3Expression.convert(to_expression)
+
+        if not isinstance(from_expression, Z3ObjectExpression):
+            raise ValueError()  # only possible when from_expression is Z3SolverExpression
```
CactusWin, line 450: Can we add a message in these two ValueError (this one and the one below)?

CactusWin (commented):
Looks good to me! Left a few comments/ nits!

rodrigodesalvobraz (changes requested):

neuralpp/symbolic/basic_expression.py
```diff
@@ -215,11 +219,25 @@ def syntactic_eq(self, other) -> bool:
                 return False
 
 
+class BasicAbelianOperation(BasicConstant, AbelianOperation):
```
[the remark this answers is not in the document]
hsiang-wu: My argument here is that `AbelianOperation` is a subclass of `Constant`: in particular, it is a subclass of "function". Similar to our design that the function `Constant(+)` is a subexpression of the function application `(+ 1 1)`, the `AbelianOperation` is a subexpression of the quantifier expression. And an `AbelianOperation(+)` is a specialize object of function `Constant(+)` in the way that we also know its identity element.

hsiang-wu:
> I hadn't seen structure_eq before, but it strikes me that this is what syntactic_eq should have been in the same place, that is, not to care about the class, only about the expression being represented. This would go along with the general philosophy of the library involving automatic conversions. Basically, the expression is still the same regardless of which class is implementing it, and conversions happen behind the scenes to ensure that.
Is there a disadvantage to making syntactic_eq do what structure_eq is doing now, and getting rid of the latter?

I agree with most of what you said here. The only advantage of `syntactic_eq` is that it is supposed to be faster. It could be a fast pass in the implementation of `structure_eq`.

neuralpp/symbolic/expression.py
```diff
@@ -164,13 +184,13 @@ def new_variable(cls, name: str, type_: ExpressionType) -> Variable:
...
     @abstractmethod
-    def new_function_application(cls, function: Expression, arguments: List[Expression]) -> FunctionApplication:
+    def new_function_application(cls, function: Expression, arguments: List[Expression]) -> Expression:
         pass
 
     @classmethod
     @abstractmethod
     def new_quantifier_expression(cls, operation: Constant, index: Variable, constrain: Expression, body: Expression,
-                                  ) -> QuantifierExpression:
+                                  ) -> Expression:
```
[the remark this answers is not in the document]
hsiang-wu: No.

https://github.com/rodrigodesalvobraz/neuralpp/pull/29/files/becc5084c2414eb21a2f88a3815b71e2565082e0#diff-12ba35f9b7c823ba859085be51887691ff64f8d9eefb76a83076b1572b2c4282R78-R82

neuralpp/symbolic/expression.py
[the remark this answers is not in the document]
hsiang-wu: There's a typo, it should be "to not necessarily return `FunctionApplication`".

neuralpp/symbolic/expression.py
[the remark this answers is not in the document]
hsiang-wu: Similarly `new_function_application()` is not guaranteed to return `FunctionApplication`. E.g., `SymPyExpression.new_function_application()` is allowed to simplify the result in some circumstances (e.g., `(and False b)`).

neuralpp/symbolic/expression.py
hsiang-wu: I agree with most of what you said here. The only advantage of syntactic_eq is that it is supposed to be faster. It could be a fast pass in the implementation of structure_eq.

neuralpp/symbolic/basic_expression.py
```diff
@@ -221,23 +221,24 @@ def syntactic_eq(self, other) -> bool:
...
+    def __init__(self, operation: Callable, identity: Expression):
         # technically, the type of identity should be element_type, but there seems no way to declare that in Python?
         self._identity = identity
+        element_type = identity.type
         super().__init__(operation, Callable[[element_type, element_type], element_type])
 
 
 def basic_add_operation(type_) -> BasicAbelianOperation:
-    return BasicAbelianOperation(operator.add, type_, 0)
+    return BasicAbelianOperation(operator.add, BasicConstant(0, type_))
```
hsiang-wu, line 235: The previous implementation was just wrong

neuralpp/symbolic/basic_expression.py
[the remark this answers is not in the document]
rodrigodesalvobraz: This is not very important but just for completeness, here's my counter-argument: just because it is a function, it does not follow that it needs to be wrapper in a Constant. In fact, we have many functions in the system (regular Python functions) that are not wrapped so.

We wrap functions as Constants when we need to create an expression containing them. Now, it is true that in the AbelianOperation, we will often use them in expressions, but not necessarily always. And by making it a Constant, we are tying the class to Expressions, creating an unnecessary dependence. It could be that we might want to use AbelianOperations separately from Expressions someday.

Anyway, given the typical usage we intend, none of this is very important. :-)

neuralpp/symbolic/expression.py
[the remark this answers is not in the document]
rodrigodesalvobraz: This is how I did it in my first implementation, and later regretted it. Today I think it is better to separate the construction of expressions and their simplification, which here are always done together. I think it is better because it is more regular: building a quantifier expressions always produces a quantifier expression, making it easier to write algorithms. It also makes typing more precise (as illustrated above). If the user wants things simplified, they can always do so.

So, if you think this is easy enough to separate, that would be good. If it is too much work, it is ok to leave it.

rodrigodesalvobraz (changes requested):

hsiang-wu:
https://github.com/rodrigodesalvobraz/neuralpp/pull/29#discussion_r930174067

It is actually not so easy to implement, and in fact it might even be impossible. Since there's some part of SymPy's simplification that we cannot turn off (even with `evaluate=False` or the global parameter trick). For example, we cannot turn off short-circuit in boolean expressions, thus `And(False, b)` is always a `Constant(False)` and never going to be a legit SymPy function application. If we really need to do the design you mentioned, we basically need to revert our design philosophy that `SymPyExpression` is backed by one SymPy object.

neuralpp/symbolic/expression.py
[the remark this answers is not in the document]
hsiang-wu: This is almost impossible to implement, as I explained in the reply below.

neuralpp/symbolic/expression.py
rodrigodesalvobraz: In that case the names could be more reflective of that. `syntactic_eq` should be `syntactic_eq_given_same_kind` and `structurally_eq` should be called `syntatically_eq`.

hsiang-wu:
renamed `structure_eq` -> `syntactic_eq`; old `syntactic_eq` -> `internal_object_eq`

CactusWin (commented):
Looks good! Lots of renaming 😅

horizon-blue (approved):
Looks pretty good to me! :)

horizon-blue:
Looks like the plan is to keep `new_quantifier_expression` as-is, as discussed [in the comment thread](https://github.com/rodrigodesalvobraz/neuralpp/pull/29#issuecomment-1195908925)? I'm going to merge the PR assuming that's the case, but it's still good to keep those discussions in mind and potentially revisit this in the future.

PR #30, opened 2022-07-28
New normalization

hsiang-wu:
A implementation of new normalization algorithm, supporting nested quantifiers.

1. created a new `GeneralNormalizer` for the new normalization; renamed old implementation to `QuantifierFreeNormalizer`
2. implemented `GeneralNormalizer.normalize()`
3. implemented a place-holding `_eliminate()` which is used in `GeneralNormalizer.normalize()`
4. supports conversion of equality/inequality `Expression`s to `bool`, so that hash can work correctly (see the comment below).


neuralpp/symbolic/context_simplifier.py
```diff
@@ -8,26 +8,12 @@
 from .parameters import sympy_evaluate
 
 
-def _is_known_to_imply(context: Context, expression: Expression) -> bool:
```
hsiang-wu, line 11: moved this function to be a method of `Context`, so a calling statement would be `context.is_known_to_imply(expression)`, which I believe is more intuitive.

neuralpp/symbolic/normalizer.py
```diff
@@ -1,108 +1,10 @@
...
-
-    @staticmethod
-    def _normalize(expression: Expression, context: Z3SolverExpression) -> Expression:
-        """ The function assumes context is satisfiable, otherwise will raise. """
-        if isinstance(expression, QuantifierExpression):
-            return Normalizer._quantifier_expression_normalize(expression, context)
-        else:
-            return Normalizer._basic_normalize(expression, context)
 
+class Normalizer(ABC):
```
hsiang-wu, line 6: changed `Normalizer` to be an interface class. The old implementation is moved to `QuantifierFreeNormalizer`

neuralpp/symbolic/quantifier_free_normalizer.py
```diff
@@ -0,0 +1,72 @@
...
+                pass  # non_constant_formulaic expression has no subexpressions
+    return None
+
+
+def first_non_constant_formula_of_expression(expression: Expression) -> Optional[Expression]:
+    """ returns the first non_constant_formula occurring in a given expression in a breadth-first search. """
+    return _bfs_first_non_constant_formula_of_expression(expression, deque())
+
+
+class QuantifierFreeNormalizer(Normalizer):
```
hsiang-wu, line 39: This is reverted to an old version of `Normalizer` before we introduce and consider the quantifier case.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -498,3 +499,26 @@ def dict(self) -> Dict[str, Any]:
     @property
     def satisfiability_is_known(self) -> bool:
         return not self._unknown
+
+
+class SymPySummation(SymPyExpression, QuantifierExpression):
```
hsiang-wu, line 504: This class is currently not used since SymPy's support of summation is not that general, so the conversion is complicated (e.g., need to split one `BasicSummation` into several `SymPySummation` due to SymPy only supports one interval in `constraint`). But it would be helpful when computing the sum in our future `_eliminate()`.

neuralpp/symbolic/expression.py
```diff
@@ -345,6 +355,22 @@ def __call__(self, *args, **kwargs) -> Expression:
                                              [arg if isinstance(arg, Expression) else self.new_constant(arg, None)
                                               for arg in args])
 
+    def __bool__(self):
```
hsiang-wu, line 358: see the comment here and `test_z3_eq_bool()` in `z3_usage_test.py`

CactusWin (approved):
Looks good to me overall! Left a few nits

neuralpp/symbolic/general_normalizer.py
```diff
@@ -0,0 +1,121 @@
...
+                        return if_then_else(condition,
+                                            _normalize(BasicQuantifierExpression(operation, index, constraint, then),
+                                                       context & condition),
+                                            _normalize(BasicQuantifierExpression(operation, index, constraint, else_),
+                                                       context & ~condition))
+                case _:
+                    return _eliminate(operation, index, constraint, normalized_body, context)
+
+
+simplifier = ContextSimplifier()
```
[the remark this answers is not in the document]
hsiang-wu: > Can we move this into _eliminate

Yes

neuralpp/test/quick_tests/symbolic/normalizer_test.py
```diff
@@ -120,11 +119,65 @@ def test_quantifier_normalizer():
...
+    #      if B>4       if B>4
+    #       /    \        /   \
+    #      f      ..   ..      f
+    #     /\                  /  \
+    #    B  Sum              C   Sum
+    #        |                    |
+    #       B*(B+C)             B*Sum(C)
+    f = BasicVariable('f', Callable[[int, int], int])
+    A = BasicVariable('A', bool)
+    B: BasicVariable = BasicVariable('B', int)
```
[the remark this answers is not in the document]
hsiang-wu: Gonna delete

neuralpp/symbolic/general_normalizer.py
```diff
@@ -0,0 +1,121 @@
...
+        return function(*arguments)
+    n_i = _normalize(arguments[i], context)
+    arguments[i] = n_i
+    return _move_down_and_normalize(function, arguments, context, i)
+
+
+def _move_down_and_normalize(function: Expression,
+                             arguments: List[Expression],
+                             context: Z3SolverExpression, i: int) -> Expression:
+    # assume i < len(arguments);
```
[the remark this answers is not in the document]
hsiang-wu: I feel these can stay in comments since they're requirements on an internal function, so these assumptions are more intended to help readers of the code understand the function (than posing requirements on the argument)? Or I think even if we were to add `assert`, they should be for debug only.

neuralpp/symbolic/general_normalizer.py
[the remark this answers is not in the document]
CactusWin: Hmmm.. you can leave it as is then. Another suggestion is that we can pass in the simplifier as an argument (and possibly pass it in to `normalize`. I'm not sure if that'll be more elegant and reduce the numbers of object created.

neuralpp/test/quick_tests/symbolic/normalizer_test.py
```diff
@@ -120,11 +119,65 @@ def test_quantifier_normalizer():
...
+    B = BasicVariable('B', int)
+    C = BasicVariable('C', int)
+    D = BasicVariable('D', int)
+    expr = f(if_then_else(A, B, C), BasicSummation(int, D, empty_context & (0 < D) & (D < 10),
+                                                   B * if_then_else(B > 4,
+                                                                    B + C,
+                                                                    BasicSummation(int, C,
+                                                                                   empty_context & (0 < C) & (C < 10),
+                                                                                   if_then_else(B < 5, C, 1)))))
+    print(normalizer.normalize(expr, empty_context))
```
horizon-blue, line 167: :) should we remove the print statement?

neuralpp/test/quick_tests/symbolic/normalizer_test.py
```diff
@@ -120,11 +119,65 @@ def test_quantifier_normalizer():
...
+    #                      / \
+    #                     C   1
+    # should be normalized to
+    #              if A
+    #           /        \
+    #      if B>4       if B>4
+    #       /    \        /   \
+    #      f      ..   ..      f
+    #     /\                  /  \
+    #    B  Sum              C   Sum
```
horizon-blue, line 153: Love the example you have here! A related question: if, instead of `B>4`, we have `B==4` here, what will happen to the `B` in the left subtree?

horizon-blue (commented):
Looks pretty great to me overall! In the future I recommend that we put code migration into its own PR (which can be merged pretty quickly) to make it easier for reviewers to focus on the more important changes :).

neuralpp/symbolic/general_normalizer.py
```diff
@@ -0,0 +1,121 @@
...
+                                 arguments=[condition, then, else_]):
+            if context.is_known_to_imply(condition):
+                return _normalize(then, context)
+            elif context.is_known_to_imply(~condition):
+                return _normalize(else_, context)
+            else:
+                return if_then_else(condition,
+                                    _normalize(then, context & condition),
+                                    _normalize(else_, context & ~condition))
+        case Expression(type=type_) if type_ == bool:
```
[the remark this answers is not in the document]
hsiang-wu: Do you mean we also have `case Expression(type=type_) if type_ == bool:` there? I believe that's not quite correct because we only want to overload `__bool__()` for `==` symbolic expressions (and `Constant(True)` & `Constant(False)` in case `==` is simplified)

neuralpp/symbolic/general_normalizer.py
```diff
@@ -0,0 +1,121 @@
...
+    arguments[i] = n_i
+    return _move_down_and_normalize(function, arguments, context, i)
+
+
+def _move_down_and_normalize(function: Expression,
+                             arguments: List[Expression],
+                             context: Z3SolverExpression, i: int) -> Expression:
+    # assume i < len(arguments);
+    # assume all {arguments[j] | j < i} does not contain if-then-else;
+    # assume arguments[i] has been normalized.
```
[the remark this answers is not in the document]
hsiang-wu: They are "if-then-else-free", which is more strict than and thus implies normalized. I'll add this explanation in comment.

neuralpp/symbolic/general_normalizer.py
```diff
@@ -0,0 +1,121 @@
...
+                    return _eliminate(operation, index, constraint, normalized_body, context)
+
+
+simplifier = ContextSimplifier()
+
+
+def _eliminate(operation: AbelianOperation, index: Variable, constraint: Context, body: Expression,
+               context: Z3SolverExpression) -> Expression:
+    """
+    Eliminates all quantifiers by doing the "summation" (in a broader sense).
```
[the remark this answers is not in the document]
hsiang-wu: Yes! This might be a better word, though I don't see it used often in the symbolic inference literature (I'm not an expert, so @rodrigodesalvobraz may want to comment on this). I use "sum"/"addition" here because it is the word for operation for [ring](https://en.wikipedia.org/wiki/Ring_(mathematics)) in group theory, from which we took inspiration for some earlier names (e.g., `AbelianOperation`).

neuralpp/test/quick_tests/symbolic/normalizer_test.py
hsiang-wu: `B` will be replaced by `4`, since [we call `ContextSimplifier.simplify()` in `_eliminate()`](https://github.com/rodrigodesalvobraz/neuralpp/pull/30/files/202e42ac2f4a8776825662798522df898b15460b#diff-54fbb81c8dfe09a753986c9a0e95c154bad9f41e90addf62fd59579ab648f1d5R83)

neuralpp/symbolic/sympy_interpreter.py
```diff
@@ -50,7 +50,10 @@ def simplify(self, expression: Expression, context: Context = TrueContext()) ->
         """
         with sympy_evaluate(True):  # To work around a bug in SymPy (see context_simplifier_test.py/test_sympy_bug).
             if not isinstance(expression, SymPyExpression):
-                expression = SymPyExpression.convert(expression)
+                try:
+                    expression = SymPyExpression.convert(expression)
+                except Exception:
+                    raise ConversionError()
```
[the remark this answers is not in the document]
hsiang-wu: TIL. Thanks for the enlightenment XD

neuralpp/symbolic/context_simplifier.py
```diff
@@ -48,7 +49,9 @@ def _simplify_pass(expression: Expression, context: Z3SolverExpression) -> Expre
             result = expression
 
         # replace boolean expressions
-        all_boolean_subexpressions = _collect_subset_expressions(result, lambda expr: expr.type == bool)
+        all_boolean_subexpressions = _collect_subset_expressions(
+            result,
+            lambda expr: expr.type == bool and not isinstance(expr, QuantifierExpression))
```
hsiang-wu, line 54: Simplifier should not look inside quantifier because of scope issue: indexed expression should not be visible outside the quantifier.

neuralpp/symbolic/general_normalizer.py
```diff
@@ -52,18 +68,14 @@ def _normalize(expression: Expression, context: Z3SolverExpression) -> Expressio
                                       BasicQuantifierExpression(operation, index, constraint & ~condition, else_)),
                             context)
                     else:
-                        return if_then_else(condition,
-                                            _normalize(BasicQuantifierExpression(operation, index, constraint, then),
-                                                       context & condition),
-                                            _normalize(BasicQuantifierExpression(operation, index, constraint, else_),
-                                                       context & ~condition))
+                        return _normalize_conditional(condition,
```
hsiang-wu, line 71: There's a bug in previous version: when `context & condition` or `context & ~condition` is not satisfiable, it should not be `_normalize()`d

neuralpp/test/quick_tests/symbolic/normalizer_test.py
```diff
@@ -205,7 +205,7 @@ def test_quantifier_normalizer():
     assert normalizer.normalize(expr, empty_context).syntactic_eq(
         if_then_else(A,
                      if_then_else(B == 4,
-                                  f(B, BasicSummation(int, D, empty_context & (0 < D) & (D < 10), 16 + 4 * C)),
+                                  f(4, BasicSummation(int, D, empty_context & (0 < D) & (D < 10), 16 + 4 * C)),
```
hsiang-wu, line 208: now `B` is simplified to `4`

neuralpp/symbolic/general_normalizer.py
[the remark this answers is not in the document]
horizon-blue: Nah I was thinking if we can rewrite it into `case Constant(value=value) if isinstance(value, bool)`, but that's not super important. :)

horizon-blue (approved):
:) Thanks for the additional test and updates to cover the `B==4` case. This looks good to me

PR #31, opened 2022-08-05
interval interface

hsiang-wu:
We (Rodrigo, Winnie, and I) had some discussions over the interval interface. We came to conclude that `Interval` should not inherit `Expression`.

The difficult part about interval is that the return type of `from_constraint(constraint)` is an if-then-else tree of intervals, which I propose an interface in this PR, using `DottedIntervals` and `ConditionalIntervals`.

### DIFF
--- a/neuralpp/symbolic/interval.py
+++ b/neuralpp/symbolic/interval.py
@@ -1,32 +1,65 @@
 from __future__ import annotations
-from abc import abstractmethod
-from typing import Iterable, Callable
-from constants import if_then_else
 
-from .expression import Variable, Expression, Context
+import builtins
+from typing import Iterable, List, Set
+
+from .expression import Variable, Expression, Context, Constant
+from .basic_expression import BasicExpression
 from .z3_expression import Z3SolverExpression
 
 
-class ClosedInterval:
+class ClosedInterval(BasicExpression):
     """ [lower_bound, upper_bound] """
     @property
-    @abstractmethod
+    def subexpressions(self) -> List[Expression]:
+        return [self.lower_bound, self.upper_bound]
+
+    def set(self, i: int, new_expression: Expression) -> Expression:
+        if i == 0:
+            return ClosedInterval(new_expression, self.upper_bound)
+        if i == 1:
+            return ClosedInterval(self.lower_bound, new_expression)
+        raise IndexError("out of scope.")
+
+    def replace(self, from_expression: Expression, to_expression: Expression) -> Expression:
+        if self.syntactic_eq(from_expression):
+            return to_expression
+        return ClosedInterval(self.lower_bound.replace(from_expression, to_expression),
+                              self.upper_bound.replace(from_expression, to_expression))
+
+    def internal_object_eq(self, other) -> bool:
+        if not isinstance(other, ClosedInterval):
+            return False
+        return self.lower_bound.internal_object_eq(other.lower_bound) and \
+               self.upper_bound.internal_object_eq(other.upper_bound)
+
+    def __init__(self, lower_bound, upper_bound):
+        super().__init__(Set)
+        self._lower_bound = lower_bound
+        self._upper_bound = upper_bound
+
+    @property
     def lower_bound(self) -> Expression:
-        pass
+        return self._lower_bound
 
     @property
-    @abstractmethod
     def upper_bound(self) -> Expression:
-        pass
+        return self._upper_bound
 
-    @abstractmethod
     def to_range(self) -> Iterable[Expression]:
-        """ if upper and lower bounds are constant, return a range that's iterable """
-        pass
+        """
+        If upper and lower bounds are constant, return a range that's iterable.
+        Otherwise, raise
+        """
+        match self.lower_bound, self.upper_bound:
+            case Constant(value=l, type=builtins.int), Constant(value=r, type=builtins.int):
+                return range(l, r)
+            case _:
+                raise TypeError("Lower and upper bounds must both be Constants!")
 
     @property
-    def length(self) -> Expression:
-        return self.upper_bound - self.lower_bound
+    def size(self) -> Expression:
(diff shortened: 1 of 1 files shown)


neuralpp/symbolic/interval.py
```diff
@@ -0,0 +1,75 @@
...
+from constants import if_then_else
+
+from .expression import Variable, Expression, Context
+from .z3_expression import Z3SolverExpression
+
+
+class ClosedInterval:
+    """ [lower_bound, upper_bound] """
+    @property
+    @abstractmethod
```
CactusWin, line 13: Should these be abstract methods?

neuralpp/symbolic/interval.py
```diff
@@ -0,0 +1,75 @@
...
+from abc import abstractmethod
+from typing import Iterable, Callable
+from constants import if_then_else
+
+from .expression import Variable, Expression, Context
+from .z3_expression import Z3SolverExpression
+
+
+class ClosedInterval:
+    """ [lower_bound, upper_bound] """
```
CactusWin, line 11: Since we're using `@property`, we might need an init function to help store self._lower_bound or self._upper_bound.

neuralpp/symbolic/interval.py
hsiang-wu: No, we can just implement this class. It's just that I want to write some legit code in a PR.

neuralpp/symbolic/interval.py
hsiang-wu: Yes. We'll implement that here. Now it's just for interface discussion.

neuralpp/symbolic/interval.py
```diff
@@ -0,0 +1,75 @@
+from __future__ import annotations
+from abc import abstractmethod
+from typing import Iterable, Callable
+from constants import if_then_else
+
+from .expression import Variable, Expression, Context
+from .z3_expression import Z3SolverExpression
+
+
+class ClosedInterval:
```
CactusWin, line 10: I think this should also inherit `GeneralIntervals', since this method also needs `from_constraint`.

hsiang-wu: The class `GeneralIntervals` is gone in the new commit

CactusWin:
I think this looks good to me!

CactusWin (approved):

neuralpp/symbolic/interval.py
```diff
@@ -0,0 +1,108 @@
...
+    def internal_object_eq(self, other) -> bool:
+        if not isinstance(other, ClosedInterval):
+            return False
+        return self.lower_bound.internal_object_eq(other.lower_bound) and \
+               self.upper_bound.internal_object_eq(other.upper_bound)
+
+    def __init__(self, lower_bound, upper_bound):
+        super().__init__(Set)
+        self._lower_bound = lower_bound
+        self._upper_bound = upper_bound
```
horizon-blue, line 39: Can we move `__init__` up to make it the first method of `ClosedInterval`?

neuralpp/symbolic/interval.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @property
+    def lower_bound(self) -> Expression:
+        return self._lower_bound
+
+    @property
+    def upper_bound(self) -> Expression:
+        return self._upper_bound
+
+    def to_range(self) -> Iterable[Expression]:
```
horizon-blue, line 49: Looks like this should be an `Iterable[int]` here since `l` and `r` are `int`s

neuralpp/symbolic/interval.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        match self.lower_bound, self.upper_bound:
+            case Constant(value=l, type=builtins.int), Constant(value=r, type=builtins.int):
+                return range(l, r)
+            case _:
+                raise TypeError("Lower and upper bounds must both be Constants!")
+
+    @property
+    def size(self) -> Expression:
+        return self.upper_bound - self.lower_bound + 1
```
horizon-blue, line 62: should we handle the case where `self.lower_bound` > `self.upper_bound`?

neuralpp/symbolic/interval.py
```diff
@@ -0,0 +1,108 @@
...
+    def set(self, i: int, new_expression: Expression) -> Expression:
+        raise NotImplementedError("TODO")
+
+    def replace(self, from_expression: Expression, to_expression: Expression) -> Expression:
+        raise NotImplementedError("TODO")
+
+    def internal_object_eq(self, other) -> bool:
+        raise NotImplementedError("TODO")
+
+    def __init__(self, interval: ClosedInterval, dots: List[Expression]):
```
horizon-blue, line 84: ditto on bringing `__init__` up

neuralpp/symbolic/interval.py
```diff
@@ -0,0 +1,108 @@
...
+    @property
+    def dots(self) -> List[Expression]:
+        return self._dots
+
+    @property
+    def interval(self) -> ClosedInterval:
+        return self._interval
+
+    @property
+    def __iter__(self) -> Iterable[Constant]:
```
horizon-blue, line 98: Why is it that we have `to_range` for `ClosedInterval` but `__iter__` for `DottedIntervals` as opposed to having `__iter__` for both?

horizon-blue (commented):
LGTM, though I'm losing a bit of context on what the `DottedIntervals` is for, so can't comment much there :).

hsiang-wu:
> LGTM, though I'm losing a bit of context on what the `DottedIntervals` is for, so can't comment much there :).

It's essentially an "interval with holes". E.g., [1,100] with holes {50,51}.

neuralpp/symbolic/interval.py
hsiang-wu: That makes sense!

PR #32, opened 2022-08-06
Super Naive and Brute-force Implementation of Intervals

CactusWin:
This is a super naive and brute-force implementation of intervals. It can only handle >, >=, <, <=. It also assumes that the context is conjoined by `and` (and not `or`). This will be improved upon! Just wanted something really simple, in case I'm blocking Xiang.

### DIFF
--- a/neuralpp/symbolic/interval.py
+++ b/neuralpp/symbolic/interval.py
@@ -1,0 +1,190 @@
+from __future__ import annotations
+
+import builtins
+import operator
+from typing import Iterable, List, Set, Tuple
+from .expression import Variable, Expression, Context, Constant, FunctionApplication
+from .basic_expression import BasicExpression
+from .z3_expression import Z3SolverExpression, Z3Expression
+
+
+class ClosedInterval(BasicExpression):
+    """ [lower_bound, upper_bound] """
+    def __init__(self, lower_bound, upper_bound):
+        super().__init__(Set)
+        self._lower_bound = lower_bound
+        self._upper_bound = upper_bound
+
+    @property
+    def lower_bound(self) -> Expression:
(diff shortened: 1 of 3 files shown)


neuralpp/symbolic/interval.py
```diff
@@ -0,0 +1,190 @@
...
+    return DottedIntervals(closed_interval, exceptions)
+
+def _extract_bound_from_constraint(
+    index: Variable,
+    constraint: Expression,
+    closed_interval: ClosedInterval,
+    exceptions: List[Expression]
+) -> Tuple(ClosedInterval, List[Expression]):
+    """
+    Helper function for from_constraints
```
rodrigodesalvobraz, line 128: I don't think a reader of this docstring will be able to completely understand what the function is doing. They would have to read the code where it is used and figure it out, which would take time.
Can you please refine it? I think the main thing to include is the relationship between output and inputs. The function returns a closed_interval and exceptions; what holds between that and the input closed_interval and exceptions, for example?

neuralpp/symbolic/interval.py
```diff
@@ -0,0 +1,190 @@
...
+        variable_index = 2
+        bound = constraint.subexpressions[1]
+    else:
+        raise ValueError("intervals is not yet ready to handle more complicated cases")
+
+    match possible_inequality:
+        case operator.ge:
+            if variable_index == 1:
+                closed_interval, exceptions = _check_and_set_bounds(0, bound, closed_interval, exceptions)
+            if variable_index == 2:
```
rodrigodesalvobraz, line 149: nit: using `elif` here would be slightly more efficient.

rodrigodesalvobraz (approved):

PR #33, opened 2022-08-08
Inverse

hsiang-wu:
We need `inverse()` method since we are computing DottedInterval and thus need to "subtract" the "sum" of dots. Also, Abelian group [by definition](https://en.wikipedia.org/wiki/Abelian_group) has "inverse elements".

Also includes the fixes to my last PR per reviews.


rodrigodesalvobraz (approved):
Great!

PR #34, opened 2022-08-09
Eliminator

hsiang-wu:
1. A simple implementation of eliminator using SymPy for symbolic computation
2. Some changes to `interval.py` to support symbolic range

### DIFF
--- a/neuralpp/symbolic/eliminator.py
+++ b/neuralpp/symbolic/eliminator.py
@@ -1,0 +1,84 @@
+from .expression import QuantifierExpression, Context, Expression, Constant, AbelianOperation, Variable, \
+    FunctionApplication
+from .basic_expression import BasicQuantifierExpression
+from .sympy_expression import SymPyExpression
+from .interval import ClosedInterval, DottedIntervals, from_constraint
+from .constants import if_then_else
+import neuralpp.symbolic.functions as functions
+from functools import reduce
+from typing import Optional, Callable
+import operator
+
+
+def _eliminate_interval(operation: AbelianOperation, index: Variable, interval: ClosedInterval, body: Expression) \
+        -> Expression:
+    # try to solve symbolically
+    result = _symbolically_eliminate(operation, index, interval, body)
+    if result is not None:
+        return result
+
+    if isinstance(interval.lower_bound, Constant) and isinstance(interval.upper_bound, Constant):
+        # iterate through the interval if we can
+        return reduce(operation,
+                      map(lambda num: body.replace(interval.index, Constant(num)), iter(interval)),
+                      operation.identity)
+
+    return BasicQuantifierExpression(operation, index, interval.to_context(index), body)
+
+
+def _repeat_n(operation: AbelianOperation, expression: Expression, size: Expression) -> Optional[Expression]:
+    if operation.value == operator.add:
+        return expression * size
+    if operation.value == operator.mul:
+        return operator.pow(expression, size)
+    return None  # TODO: expand this
+
+
+def _symbolically_eliminate(operation: AbelianOperation, index: Variable, interval: ClosedInterval, body: Expression) \
+        -> Optional[Expression]:
+    if operation.value == operator.add:
+        if (result := Eliminator.symbolic_sum(body, index, interval)) is not None:
+            return result
+
+    if isinstance(body, Constant):
+        # repeat addition: multiplication, repeat multiplication: power
+        return _repeat_n(operation, body, interval.size)
+    return None
+
+
+def _map_leaves(conditional_intervals: Expression,
+                function: Callable[[DottedIntervals], Expression]) -> Expression:
+    """
+    @param conditional_intervals: expect an if-then-else tree with leaves being DottedIntervals
+    @param function: map function
+    @return: an Expression with each DottedIntervals i mapped to f(i)
+    """
+    match conditional_intervals:
+        case FunctionApplication(function=Constant(value=functions.conditional), arguments=[if_, then, else_]):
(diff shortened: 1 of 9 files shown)


neuralpp/symbolic/interval.py
```diff
@@ -48,24 +49,25 @@ def internal_object_eq(self, other) -> bool:
...
-
     def __iter__(self) -> Iterable[int]:
         """
         If upper and lower bounds are constant, return a range that's iterable.
         Otherwise, raise
         """
         match self.lower_bound, self.upper_bound:
             case Constant(value=l, type=builtins.int), Constant(value=r, type=builtins.int):
-                return range(l, r)
+                return iter(range(l, r))
```
hsiang-wu, line 59: bug fix

neuralpp/symbolic/interval.py
```diff
@@ -48,24 +49,25 @@ def internal_object_eq(self, other) -> bool:
...
 
     @property
     def size(self) -> Expression:
+        if self.lower_bound > self.upper_bound:
+            raise AttributeError(f'[{self.lower_bound},{self.upper_bound}] is an empty interval.')
         return self.upper_bound - self.lower_bound + 1
 
     def to_context(self, index: Variable) -> Context:
-        result = Z3SolverExpression() & index >= self.lower_bound & index <= self.upper_bound
+        result = Z3SolverExpression() & (index >= self.lower_bound) & (index <= self.upper_bound)
```
hsiang-wu, line 70: bug fix

neuralpp/symbolic/interval.py
```diff
@@ -167,28 +170,29 @@ def _extract_bound_from_constraint(
                 closed_interval, exceptions = _check_and_set_bounds(0, bound, closed_interval, exceptions)
         case operator.gt:
             if variable_index == 1:
-                bound = Z3Expression.new_constant(bound.value + 1)
+                bound = _simplifier.simplify(bound + 1)
```
hsiang-wu, line 173: changed this so `bound` does not have to be a `Constant`

neuralpp/symbolic/sympy_expression.py
```diff
@@ -152,6 +157,24 @@ def __init__(self, sympy_object: sympy.Basic, expression_type: ExpressionType,
         self._sympy_object = sympy_object
         self._type_dict = type_dict
 
+    @staticmethod
+    def symbolic_sum(body: Expression, index: Variable, lower_bound: Expression, upper_bound: Expression) \
```
hsiang-wu, line 161: Use SymPy to do symbolic computation

neuralpp/symbolic/eliminator.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def _eliminate_interval(operation: AbelianOperation, index: Variable, interval: ClosedInterval, body: Expression) \
+        -> Expression:
+    # try to solve symbolically
+    result = _symbolically_eliminate(operation, index, interval, body)
+    if result is not None:
+        return result
+
+    # TODO: enable this
```
hsiang-wu, line 20: temporarily comment out this because for our current scope, symbolic eliminate should always work

neuralpp/symbolic/eliminator.py
```diff
@@ -0,0 +1,85 @@
...
+        if (result := Eliminator.symbolic_sum(body, index, interval)) is not None:
+            return result
+
+    if isinstance(body, Constant):
+        # repeat addition: multiplication, repeat multiplication: power
+        return _repeat_n(operation, body, interval.size)
+    return None
+
+
+def _map_leaves(conditional_intervals: Expression,
```
rodrigodesalvobraz, line 50: This seems like a generic enough function (it doesn't seem to need to be about `DottedIntervals` in the leaves, it could just `Expression`) that could be placed in a util file.

rodrigodesalvobraz (changes requested):
Great! Just a few minor comments.

neuralpp/symbolic/eliminator.py
```diff
@@ -0,0 +1,85 @@
...
+        return result
+
+    # TODO: enable this
+    # if isinstance(interval.lower_bound, Constant) and isinstance(interval.upper_bound, Constant):
+    #     # iterate through the interval if we can
+    #     return reduce(operation,
+    #                   map(lambda num: body.replace(interval.index, Constant(num)), iter(interval)),
+    #                   operation.identity)
+
+    return BasicQuantifierExpression(operation, index, interval.to_context(index), body)
```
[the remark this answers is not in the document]
hsiang-wu: Kind of. Though that's not exactly the same, since `from_constraint` (supposedly if not currently) returns several intervals in an if-then-else tree, and here we only convert to context one interval (or `DottedInterval`).

neuralpp/symbolic/eliminator.py
```diff
@@ -0,0 +1,85 @@
...
+    return None  # TODO: expand this
+
+
+def _symbolically_eliminate(operation: AbelianOperation, index: Variable, interval: ClosedInterval, body: Expression) \
+        -> Optional[Expression]:
+    if operation.value == operator.add:
+        if (result := Eliminator.symbolic_sum(body, index, interval)) is not None:
+            return result
+
+    if isinstance(body, Constant):
```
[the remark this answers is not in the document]
hsiang-wu: `symbolic_sum` only deals with summation, this chunk of code can deal with other operations.

neuralpp/symbolic/eliminator.py
hsiang-wu: Yes

PR #35, opened 2022-08-09
add support for integration

hsiang-wu:
The motivation of this PR is to support integration/integrals. Integral is a very similar "quantifier operation" to sum, in the way that they behave exactly the same in normalization algorithm except the final "elimination" (in elimination, integrals integrate, while sums sum). In this PR, I added a property `is_integral` to `QuantifierExpression`, to represent Integral. I am not exactly sure that this is the most elegant design, but this is a rather short path to the goal given our current code base.

### DIFF
--- a/neuralpp/symbolic/basic_expression.py
+++ b/neuralpp/symbolic/basic_expression.py
@@ -77,11 +77,12 @@ class BasicExpression(Expression, ABC):
     @classmethod
     def new_quantifier_expression(cls,
                                   operation: AbelianOperation, index: Variable, constraint: Context, body: Expression,
+                                  is_integral: bool,
                                   ) -> Expression:
         if constraint.satisfiability_is_known and constraint.unsatisfiable:
             return operation.identity
         else:
-            return BasicQuantifierExpression(operation, index, constraint, body)
+            return BasicQuantifierExpression(operation, index, constraint, body, is_integral)
 
 
 class BasicAtomicExpression(BasicExpression, AtomicExpression, ABC):
@@ -219,11 +220,13 @@ class BasicFunctionApplication(BasicExpression, FunctionApplication):
 
 
 class BasicQuantifierExpression(QuantifierExpression, BasicExpression):
-    def __init__(self, operation: AbelianOperation, index: Variable, constraint: Context, body: Expression):
+    def __init__(self, operation: AbelianOperation, index: Variable, constraint: Context, body: Expression,
+                 is_integral: bool):
         self._operation = operation
         self._index = index
         self._constraint = constraint
         self._body = body
+        self._is_integral = is_integral
         argument_types, return_type = get_args(operation.type)
         if len(argument_types) != 2 or not (argument_types[0] == argument_types [1] == return_type):
             raise ValueError(f"Wrong operation type {operation.type}.")
@@ -231,6 +234,10 @@ class BasicQuantifierExpression(QuantifierExpression, BasicExpression):
 
     def __hash__(self):
         return hash(tuple(self.subexpressions))
+
+    @property
+    def is_integral(self) -> bool:
+        return self._is_integral
 
     @property
     def operation(self) -> Constant:
@@ -254,8 +261,9 @@ class BasicQuantifierExpression(QuantifierExpression, BasicExpression):
 
     def internal_object_eq(self, other) -> bool:
         match other:
-            case BasicQuantifierExpression(subexpressions=other_subexpressions):
-                return all(lhs.internal_object_eq(rhs) for lhs, rhs in zip(self.subexpressions, other_subexpressions))
+            case BasicQuantifierExpression(subexpressions=other_subexpressions, is_integral=other_is_integral):
+                return self.is_integral == other_is_integral and \
+                       all(lhs.internal_object_eq(rhs) for lhs, rhs in zip(self.subexpressions, other_subexpressions))
             case _:
                 return False
 
@@ -281,10 +289,14 @@ def basic_add_operation(type_) -> BasicAbelianOperation:
     return BasicAbelianOperation(operator.add, BasicConstant(0, type_), operator.neg)
 
 
-class BasicSummation(BasicQuantifierExpression):
-    def __init__(self, type_: type, index: Variable, constraint: Context, body: Expression):
-        """
-        Expect type_ to be the argument type/return type of the summation.
-        E.g., the type_ of Summation{i in [0,100]}(i) should be `int`.
-        """
-        super().__init__(basic_add_operation(type_), index, constraint, body)
+def BasicSummation(type_: type, index: Variable, constraint: Context, body: Expression) -> BasicQuantifierExpression:
+    """
+    Expect type_ to be the argument type/return type of the summation.
+    E.g., the type_ of Summation{i in [0,100]}(i) should be `int`.
+    """
+    return BasicQuantifierExpression(basic_add_operation(type_), index, constraint, body, False)
+
+
+def BasicIntegral(index: Variable, constraint: Context, body: Expression) -> BasicQuantifierExpression:
+    # basic_add_operation.element_type is redundant here
+    return BasicQuantifierExpression(basic_add_operation(int), index, constraint, body, True)
--- a/neuralpp/symbolic/eliminator.py
+++ b/neuralpp/symbolic/eliminator.py
@@ -8,10 +8,11 @@ from typing import Optional
 import operator
 
 
-def _eliminate_interval(operation: AbelianOperation, index: Variable, interval: ClosedInterval, body: Expression) \
+def _eliminate_interval(operation: AbelianOperation, index: Variable, interval: ClosedInterval, body: Expression,
+                        is_integral: bool) \
         -> Expression:
     # try to solve symbolically
-    result = _symbolically_eliminate(operation, index, interval, body)
+    result = _symbolically_eliminate(operation, index, interval, body, is_integral)
     if result is not None:
         return result
 
@@ -21,8 +22,7 @@ def _eliminate_interval(operation: AbelianOperation, index: Variable, interval: 
     #     return reduce(operation,
     #                   map(lambda num: body.replace(interval.index, Constant(num)), iter(interval)),
     #                   operation.identity)
-
-    return BasicQuantifierExpression(operation, index, interval.to_context(index), body)
+    return BasicQuantifierExpression(operation, index, interval.to_context(index), body, is_integral)
 
 
 def _repeat_n(operation: AbelianOperation, expression: Expression, size: Expression) -> Optional[Expression]:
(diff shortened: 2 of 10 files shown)


neuralpp/symbolic/eliminator.py
```diff
@@ -33,11 +33,16 @@ def _repeat_n(operation: AbelianOperation, expression: Expression, size: Express
...
+                            is_integral: bool) \
         -> Optional[Expression]:
     if operation.value == operator.add:
-        if (result := Eliminator.symbolic_sum(body, index, interval)) is not None:
-            return result
+        if not is_integral:
+            if (result := Eliminator.symbolic_sum(body, index, interval)) is not None:
+                return result
+        else:
+            if (result := Eliminator.symbolic_integral(body, index, interval)) is not None:
```
hsiang-wu, line 44: Add a branch for symbolic computation of integrals

neuralpp/symbolic/sympy_expression.py
```diff
@@ -270,14 +293,16 @@ def internal_object_eq(self, other) -> bool:
...
+    def from_sympy_object(sympy_object: sympy.Basic, type_dict: Dict[sympy.Basic, ExpressionType]) -> SymPyExpression:
         # Here we just try to find a type of expression for sympy object.
         if isinstance(sympy_object, sympy.Symbol):
             return SymPyVariable(sympy_object, type_dict[sympy_object])
         elif _is_sympy_value(sympy_object):
             return SymPyConstant(sympy_object, _infer_sympy_object_type(sympy_object, type_dict))
         elif _is_sympy_sum(sympy_object):
             return SymPySummation(sympy_object, type_dict)
+        elif _is_sympy_integral(sympy_object):
+            raise NotImplementedError("expect sympy to eliminate integral sign. TODO")
```
hsiang-wu, line 305: We don't need a `SymPyIntegral` class for now, if we don't need to represent integral symbolically. It is also easy to add support for `SymPyIntegral`.

neuralpp/test/quick_tests/symbolic/normalizer_test.py
```diff
@@ -115,27 +115,36 @@ def test_quantifier_normalizer():
...
     assert normalizer.normalize(sum_, empty_context).syntactic_eq(
         if_then_else(j > 5,
-                     BasicSummation(int, i, empty_context & (j < i), i + j),
-                     BasicSummation(int, i, empty_context & (j < i), i)))
+                     SymPyExpression.from_sympy_object(
+                         - jj ** 2 / 2 - jj * (jj - kk + 1) - jj / 2 + kk ** 2 / 2 - kk / 2,
+                         {jj: int, kk: int}),
+                     SymPyExpression.from_sympy_object(- jj ** 2 / 2 - jj / 2 + kk ** 2 / 2 - kk / 2,
+                                                       {jj: int, kk: int}),
+                     ))
```
hsiang-wu, line 134: This test file went through a lot of changes since we can now actually eliminate quantifiers (for simple intervals).

neuralpp/symbolic/interval.py
```diff
@@ -104,10 +108,11 @@ def __iter__(self) -> Iterable[int]:
         raise NotImplementedError("TODO")
 
 
-def from_constraint(index: Variable, constraint: Context) -> Expression:
+def from_constraint(index: Variable, constraint: Context, context: Context, is_integral: bool) -> Expression:
```
hsiang-wu, line 111: I believe this should be the correct interface. Instead of `from_constraint(index, constraint & context, is_integral)`,
we should call `from_constraint(index, constraint, context, is_integral)` and provide the context separately. This also enables more test cases given our current interval implementation.

neuralpp/symbolic/interval.py
```diff
@@ -170,17 +179,21 @@ def _extract_bound_from_constraint(
                 closed_interval, exceptions = _check_and_set_bounds(0, bound, closed_interval, exceptions)
         case operator.gt:
             if variable_index == 1:
-                bound = _simplifier.simplify(bound + 1)
+                if not is_integral:
```
hsiang-wu, line 182: Don't round when it's integral.

neuralpp/test/quick_tests/symbolic/normalizer_test.py
```diff
@@ -176,22 +185,17 @@ def test_quantifier_normalizer():
                                                                     BasicSummation(int, C,
                                                                                    empty_context & (0 < C) & (C < 10),
                                                                                    if_then_else(B < 5, C, 1)))))
+    BB, CC = sympy.symbols('B C')
+    product = SymPyExpression.from_sympy_object(9 * BB * (BB + CC), {BB: int, CC: int})
```
hsiang-wu, line 189: I called `from_sympy_object` here because it's not same as `9 * B * (B+C)`, since SymPy uses a 3-argument multiplication internally. Our overloaded `*` is a 2-argument multiplication.

neuralpp/symbolic/interval.py
CactusWin: Clarifying question: what is the difference between constraint and context?

CactusWin (commented):
Great work! Added a few nits and a clarifying question :)

neuralpp/test/quick_tests/symbolic/normalizer_test.py
```diff
@@ -115,27 +115,36 @@ def test_quantifier_normalizer():
...
+    # 6 + 7 + 8 + 9 + 5 * 4 = 50
+    assert normalizer.normalize(sum_, context).syntactic_eq(BasicConstant(50))  # SymPy switched 5 and i
 
     context = empty_context & (j == 10)
     assert normalizer.normalize(sum_, context).syntactic_eq(BasicConstant(0, int))
 
-    sum_ = BasicSummation(int, i, empty_context & (j < i), if_then_else(j > 5, i + j, i))
+    k = BasicVariable('k', int)
+    jj, kk = sympy.symbols('j k')
+    sum_ = BasicSummation(int, i, empty_context & (j < i) & (i < k), if_then_else(j > 5, i + j, i))
```
rodrigodesalvobraz, line 126: nit: using `empty_context & ` to create contexts from expressions seems a little unnatural. Why not create something like `Context.from_expression(expr)`?

neuralpp/test/quick_tests/symbolic/normalizer_test.py
```diff
@@ -214,11 +218,65 @@ def test_quantifier_normalizer():
...
+    #                      / \
+    #                     C   1
+    # should be normalized to
+    #              if A
+    #           /        \
+    #      if B>4       if B>4
+    #       /    \        /   \
+    #      *      ..   ..      * (= C * 10 * 50 = 500 * B * C)
+    #     /\                  /  \
+    #    B  Integral(D)      C   Integral(D)    (= *10)
```
rodrigodesalvobraz, line 259: I am having a hard time matching this diagram and the actual expressions in the program. For example, the diagram shows `Integral(C)` in the last line 261 without a body, and in that same line we see `50 * B` but the code has `500*B` instead. Line 259 includes an expression `*10` (is this a product of a single number?).

rodrigodesalvobraz (changes requested):
I wonder if you really need to pass `is_integral` around a lot like that, or if it would be enough to have a property checking whether the index is continuous or discrete.

neuralpp/symbolic/interval.py
hsiang-wu: `constraint` is part of the quantifier expression. `context` is "current context in the if-then-else tree" so e.g. in `if A then (if B then C else D) else E`, the context of C is `A & B`.

neuralpp/symbolic/eliminator.py
```diff
@@ -48,22 +53,26 @@ def _symbolically_eliminate(operation: AbelianOperation, index: Variable, interv
 class Eliminator:
     @staticmethod
     def eliminate(operation: AbelianOperation, index: Variable, constraint: Context, body: Expression,
-                  context: Context) -> Expression:
+                  is_integral: bool, context: Context) -> Expression:
```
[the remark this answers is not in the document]
CactusWin: I'm referring to the ordering of the parameters, rather than adding `is_integral` inside of `Context`

neuralpp/symbolic/basic_expression.py
```diff
@@ -281,10 +289,14 @@ def basic_add_operation(type_) -> BasicAbelianOperation:
...
-        super().__init__(basic_add_operation(type_), index, constraint, body)
+def BasicSummation(type_: type, index: Variable, constraint: Context, body: Expression) -> BasicQuantifierExpression:
+    """
+    Expect type_ to be the argument type/return type of the summation.
+    E.g., the type_ of Summation{i in [0,100]}(i) should be `int`.
+    """
+    return BasicQuantifierExpression(basic_add_operation(type_), index, constraint, body, False)
+
+
+def BasicIntegral(index: Variable, constraint: Context, body: Expression) -> BasicQuantifierExpression:
```
horizon-blue, line 300: N00b question: why are we making `BasicSummation` and `BasicIntegral` functions instead of having them as classes now?

neuralpp/symbolic/basic_expression.py
```diff
@@ -219,11 +220,13 @@ def internal_object_eq(self, other) -> bool:
 
 
 class BasicQuantifierExpression(QuantifierExpression, BasicExpression):
-    def __init__(self, operation: AbelianOperation, index: Variable, constraint: Context, body: Expression):
+    def __init__(self, operation: AbelianOperation, index: Variable, constraint: Context, body: Expression,
```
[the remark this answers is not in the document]
hsiang-wu: After discussion with @rodrigodesalvobraz, we decided to keep it, since it's a hack anyway. However, we should revisit the design of integral later.

neuralpp/symbolic/basic_expression.py
hsiang-wu: I should have explained this in the comment. Thanks for asking.

Because we're constantly re-creating new `BasicQuantifierExpression` by `new_quantifier_expression` (called by `replace()` which is used everywhere in simplification and normalization). And these new `BasicQuantifierExpression`s are not `BasicSummation` or `BasicIntegral`. So it's kinda weird to me that we have two notions (`BasicSummation` and `BasicQuantifierExpression` whose operation is `operator.add`) at the same time representing the same thing, and sometimes the former get transformed into the latter. So I just decided to make `BasicSummation` and `BasicIntegral` functions.

horizon-blue: Gotcha, thanks for the clarification! I was asking because I felt it would be very convenient to use the class to check if an expression is integral, but I agree that it can be tricky to deal with summations and integrals that have the base quantifier type.

neuralpp/test/quick_tests/symbolic/normalizer_test.py
hsiang-wu: I've changed to that.

neuralpp/test/quick_tests/symbolic/normalizer_test.py
hsiang-wu: Added comments explaining that

rodrigodesalvobraz (approved):

PR #37, opened 2022-08-11
lazy normalizer

hsiang-wu:
Lazy normalizer: do not normalize other than QuantifierExpression. Could be faster.


rodrigodesalvobraz (approved):
Cool!

PR #40, opened 2022-08-17
Incremental anytime exact belief propagation

shannyang:
Implements and tests incremental AEBP!

### DIFF
--- a/neuralpp/experiments/experimental_inference/exact_belief_propagation.py
+++ b/neuralpp/experiments/experimental_inference/exact_belief_propagation.py
@@ -1,4 +1,7 @@
-from neuralpp.experiments.experimental_inference.graph_analysis import LazyFactorSpanningTree, FactorGraph, FactorTree
+from neuralpp.experiments.experimental_inference.graph_analysis import LazyFactorSpanningTree, FactorGraph, FactorTree, \
+    PartialFactorSpanningTree
+from neuralpp.experiments.experimental_inference.graph_computation import ExpansionValueComputation, \
+    PartialTreeComputation
 from neuralpp.inference.graphical_model.representation.factor.product_factor import ProductFactor
 from neuralpp.util import util
 
@@ -30,3 +33,64 @@ class BeliefPropagation:
 class ExactBeliefPropagation(BeliefPropagation):
     def __init__(self, factors, query):
         super().__init__(LazyFactorSpanningTree(FactorGraph(factors), query))
+
+
+class IncrementalAnytimeBeliefPropagation(PartialTreeComputation):
+
+    def __init__(self, partial_tree, full_tree, approximation_fn, expansion_fn):
+        super().__init__(partial_tree, full_tree)
+        self.approximation = approximation_fn
+        self.expansion = ExpansionValueComputation(partial_tree, full_tree, expansion_fn)
+        self.compute_result_dict(partial_tree.root)
+
+    @staticmethod
+    def from_factors(factors, query, approximation_fn, expansion_fn):
+        """
+        Simple utility to start AnytimeBeliefPropagation on a partial tree containing just the root.
+        """
+        full_tree = LazyFactorSpanningTree(FactorGraph(factors), query)
+        partial_tree = PartialFactorSpanningTree(FactorGraph(factors), query)
+        return IncrementalAnytimeBeliefPropagation(partial_tree, full_tree, approximation_fn, expansion_fn)
+
+    def compute_result_dict(self, node):
+        self[node] = None
+        self.message_from(node)
+
+    def update_value(self, target_node, changed_child, child_value):
+        self[target_node] = None
+        self.message_from(target_node)
+
+    def run(self):
+        return self[self.partial_tree.root].normalize()
+
+    def message_from(self, node):
+        if node not in self.partial_tree:
+            return self.approximation(node, self.partial_tree, self.full_tree)
+        cached = self.result_dict.get(id(node), None)
+        if cached is not None:
+            return cached
+        product_at_node = self.product_at(node)
+        vars_summed_out = self.variables_summed_out_at(node, product_at_node.variables)
+        self[node] = product_at_node ^ vars_summed_out
+        return self[node]
+
+    def product_at(self, node):
+        incoming_messages = [self.message_from(n) for n in self.full_tree.children(node)]
+        return ProductFactor.multiply(self.partial_tree.factor_at(node) + incoming_messages)
+
+    def variables_summed_out_at(self, node, all_variables_in_product_at_node):
+        return util.subtract(
+            all_variables_in_product_at_node,
+            self.partial_tree.external_variables(node)
+        )
+
+    def expand_partial_tree_and_recompute(self, expansion_root):
+        expanded_node = self.expansion[expansion_root]
+        if expanded_node is None:
+            return
+        self.expansion.expand_partial_tree_and_recompute(expansion_root)
+        self.recompute_and_propagate_result_to_parents(expanded_node[0])
+
+    def is_complete(self):
+        return self.expansion.is_complete()
+
--- a/neuralpp/experiments/experimental_inference/exact_belief_propagation.py
+++ b/neuralpp/experiments/experimental_inference/exact_belief_propagation.py
@@ -1,6 +1,8 @@
+from collections import namedtuple
+
 from neuralpp.experiments.experimental_inference.graph_analysis import LazyFactorSpanningTree, FactorGraph, FactorTree, \
-    PartialFactorSpanningTree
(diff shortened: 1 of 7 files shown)


neuralpp/experiments/experimental_inference/graph_computation.py
```diff
@@ -0,0 +1,110 @@
...
+        self.compute_result_dict(node)
+        node_result = self[node]
+        target_node = self.partial_tree.parent(node)
+        while target_node is not None:
+            self.update_value(target_node, node, node_result)
+            node = target_node
+            target_node = self.partial_tree.parent(target_node)
+
+
+class ExpansionValueComputation(PartialTreeComputation):
```
rodrigodesalvobraz, line 47: I think we need to rethink the design of this expansion tree; it seems to be dong several different things at once. Let's talk about it in our meeting today.

rodrigodesalvobraz (changes requested):
Great stuff, Shannon! There was a lot of content here so the review is somewhat detailed. It would be great if you can go over it before our meeting today! Thanks!

neuralpp/experiments/experimental_inference/exact_belief_propagation.py
```diff
@@ -31,3 +33,64 @@ def variables_summed_out_at(self, node, all_variables_in_product_at_node):
...
+        """
+        full_tree = LazyFactorSpanningTree(FactorGraph(factors), query)
+        partial_tree = PartialFactorSpanningTree(FactorGraph(factors), query)
+        return IncrementalAnytimeBeliefPropagation(partial_tree, full_tree, approximation_fn, expansion_fn)
+
+    def compute_result_dict(self, node):
+        self[node] = None
+        self.message_from(node)
+
+    def update_value(self, target_node, changed_child, child_value):
```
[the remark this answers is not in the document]
shannyang: Yes, both methods are overrides from the TreeComputation. For belief propagation they're identical because it doesn't matter which child is changed or what the child's value is. On the other hand, the expansion computation cares about a new child_value and whether it's higher than the ancestor's existing value.

`changed_child` isn't actually used by either computation - I added it anticipating that some tree computations might want it, but could just take that out.

neuralpp/experiments/experimental_inference/graph_computation.py
```diff
@@ -31,80 +29,60 @@ def __call__(self, item):
...
-                if self[child] is not None
-            ]
-            update_candidates.append(child_value)
-            child_value = argmax(
-                update_candidates,
-                lambda node_value_pair: node_value_pair[1],
-            )
-        self[target_node] = child_value
 
     def expand_partial_tree_and_recompute(self, expansion_root):
```
rodrigodesalvobraz, line 77: Note that `TreeComputation.invalidate` does not assume this is a partial tree to be expanded. It just invalidates a value for whatever reason. This makes `TreeComputation` more generic and maybe useful for situation in which we don't have an expanding partial tree, but something else.

I think it would be good to bring the same lack of assumption to `ExpansionValueComputation`. Currently it is offering `expand_partial_tree_and_recompute` to expand the tree, but this requires the class to know that it is operating on a partial tree and to know what is the full tree and so on. Note that this method is the only place where `expand_partial_tree_and_recompute` effectively `full_tree`. In the other methods, it passes `full_tree` to `expansion_value_function`, but we could avoid that by having the expansion function be created for an already specific full tree, externally, and only receive a node and the (partial) tree as arguments. This would "free" `ExpansionValueComputation` from the assumption that it is operating on a partial tree, and it would not have to expand the tree. It would simply compute the expansion value.

This would have the added benefit of moving the job of expanding the tree to where I think it belongs more, which is  `AnytimeExactBeliefPropagation`, because that's where intuitively the action of expanding a partial tree for better approximations happens.

rodrigodesalvobraz (changes requested):
Hi Shannon. I finally finished this review. It details some points about making `TreeComputation` easier to understand, hopefully this suggestions will make sense. Thanks!

neuralpp/experiments/experimental_inference/graph_computation.py
shannyang: Tried to rework it in [this commit](https://github.com/rodrigodesalvobraz/neuralpp/commit/00aec9e565c7dd86153d5293bac0929b7cd59241), let me know what you think! The main issue was that the expansion was working on an implicitly larger partial tree (the AEBP partial tree + boundaries). I introduced `PartialExpansionTree` to represent this tree explicitly and think it's neater on the computation side, but I'm not sure if it's ok to have so many different tree classes floating around.

neuralpp/experiments/experimental_inference/graph_computation.py
```diff
@@ -1,88 +1,74 @@
...
-        assert(parent is not None)
-        self.tree.add_edge(parent, expand_to_node)
-
-        # Compute available children's expansion results and update value of expanded node
-        # to the highest available child result, then propagate through ancestors:
-        self.recompute_and_propagate_result_to_ancestors(expand_to_node)
-
-    def is_complete(self):
-        return all(self.result_dict[k] is None for k in self.result_dict)
+            candidates = [self[child] for child in children if self[child] is not None]
```
rodrigodesalvobraz, line 73: What happens if this list turns out to be empty?

neuralpp/experiments/experimental_inference/graph_computation.py
```diff
@@ -1,88 +1,74 @@
...
+        """
+        while node is not None:
+            if node in self:
+                del self.result_dict[id(node)]
+            node = self.tree.parent(node)
+
+
+class MaximumLeafValueComputation(TreeComputation):
+
+    def __init__(self, tree: Tree, leaf_value_function, argmax_key=None):
```
rodrigodesalvobraz, line 62: Is `argmax_key` really optional? It seems there will be an error if it is `None`.

neuralpp/experiments/experimental_inference/exact_belief_propagation.py
```diff
@@ -35,18 +37,24 @@ def __init__(self, factors, query):
         super().__init__(LazyFactorSpanningTree(FactorGraph(factors), query))
 
 
+Expansion = namedtuple("Expansion", "node expansion_value")
+
+
 class AnytimeExactBeliefPropagation(TreeComputation):
```
rodrigodesalvobraz, line 43: Like I mentioned in our meeting, `AnytimeExactBeliefPropagation` is taking care of both probabilistic inference and managing the mechanics of expanding a `TreeComputation` and updating it. Would it be possible to define a `PartialTreeComputation` class that subclasses `TreeComputation` and takes care of expanding/updating values? Then `AnytimeExactBeliefPropagation` would just define the message computations.

neuralpp/experiments/experimental_inference/exact_belief_propagation.py
```diff
@@ -92,9 +93,13 @@ def expand_partial_tree_and_recompute(self, expansion_root):
         potential_expansion = self.expansion[expansion_root]
         if potential_expansion is None:
             return
-        self.expansion.expand_partial_tree_and_recompute(expansion_root)
-        self.recompute_and_propagate_result_to_ancestors(potential_expansion.node)
+        expand_to_node = potential_expansion.node
+        parent = self.full_tree.parent(expand_to_node)
+
+        self.tree.add_edge(parent, expand_to_node)
```
rodrigodesalvobraz, line 99: If we have a super class `PartialTreeComputation`, then adding an edge to it would be all that's needed at this point. The value updating mechanics would happen there.

neuralpp/experiments/experimental_inference/exact_belief_propagation.py
```diff
@@ -92,9 +93,13 @@ def expand_partial_tree_and_recompute(self, expansion_root):
...
+        self.expansion.update_value(expand_to_node)
```
rodrigodesalvobraz, line 100: As mentioned in the meeting, wouldn't be better if `PartialTreeComputation` simply invalidates a node and its parent's values when an edge is edged, but waited until a value is actually requested to compute the new values?

rodrigodesalvobraz (changes requested):

neuralpp/experiments/experimental_inference/graph_computation.py
shannyang: Yes, as of 3.8: https://docs.python.org/3/library/functions.html#max

As far as I can tell it goes by identity/usual max function if the key is None.

neuralpp/experiments/experimental_inference/graph_computation.py
shannyang: We rewrote `argmax` to return `None` on an empty collection, so this should be fine.

rodrigodesalvobraz (approved):

PR #45, opened 2022-08-30
Fixed or Comment Out Test

CactusWin:
All test are passing now. I'll fix and debug the commented out/ skip tests in a future pull request.

`pytest test/quick_tests/` => 370 passed, 1 skipped, 10 warnings


rodrigodesalvobraz:
Great, all tests passing now, thanks!

hsiang-wu:
Thanks for doing this! The (commented out) normalizer tests should be correct, but I changed the code a bit and so the results are in a slightly different form, and these results are hard to write by hand...

PR #48, opened 2022-09-09
Poly implementation cleanup

hsiang-wu:
Created a new class `SymPyPoly` to wrap `sympy.Poly`. It is not a subclass of `FunctionApplication`, because the polynomial can also be a `Constant` (e.g., `1`), or a `Variable` (e.g., `x`).

### DIFF
--- a/neuralpp/symbolic/basic_expression.py
+++ b/neuralpp/symbolic/basic_expression.py
@@ -55,7 +55,7 @@ class BasicAtomicExpression(BasicExpression, AtomicExpression, ABC):
     def __init__(self, atom: Any, expression_type: Optional[ExpressionType] = None):
         if expression_type is None and not isinstance(atom, ExpressionType):
             # try to infer type for atom
-            if isinstance(atom, Callable):
+            if callable(atom):
                 expression_type = infer_python_callable_type(atom)
             else:
                 expression_type = type(atom)
--- a/neuralpp/symbolic/expression.py
+++ b/neuralpp/symbolic/expression.py
@@ -19,6 +19,10 @@ class Expression(ABC):
         self._type = expression_type
 
     @property
+    def is_polynomial(self) -> bool:
+        return False
+
+    @property
     @abstractmethod
     def subexpressions(self) -> List[Expression]:
         """
@@ -108,6 +112,7 @@ class Expression(ABC):
         E.g, a Z3Expression of "a + b" is not considered equal to a SymPyExpression of "a + b" by internal_object_eq(),
         but is considered equal to it by syntactic_eq().
         """
+        from neuralpp.symbolic.sympy_expression import SymPyPoly
         if self.internal_object_eq(other):
             return True
 
@@ -117,22 +122,11 @@ class Expression(ABC):
             ), AtomicExpression(
                 base_type=other_base_type, atom=other_atom, type=_
             ):
-                # TODO: fix this
+                # TODO: fix this. If SymPyExpression typing system is fixed, `self_type == other_type` should work
                 # return self_base_type == other_base_type and self_type == other_type and self_atom == other_atom
                 return self_base_type == other_base_type and self_atom == other_atom
-            case (  # TODO: remove this case, which is taking care of semantic rather than syntactic equality
-                    # At the moment the treatment of Polynomials depends on this because they are represented
-                    # as applications of the identity function. Once we represent them in their own class,
-                    # we can remove this case.
-                     FunctionApplication(function=Constant(value=functions.identity), arguments=identity_arguments),
-                     the_other
-                 ) | \
-                 (
-                     the_other,
-                     FunctionApplication(function=Constant(value=functions.identity), arguments=identity_arguments)
-                 ):
-                assert len(identity_arguments) == 1
-                return identity_arguments[0].syntactic_eq(the_other)
+            case (SymPyPoly(poly=poly), the_other) | (the_other, SymPyPoly(poly=poly)):
+                return poly.syntactic_eq(the_other)
             case (
                      FunctionApplication(subexpressions=self_subexpressions),
                      FunctionApplication(subexpressions=other_subexpressions),
@@ -437,10 +431,6 @@ class Variable(AtomicExpression, ABC):
 
 class FunctionApplication(Expression, ABC):
     __match_args__ = ("function", "arguments", "number_of_arguments")
-
-    @property
-    def is_polynomial(self) -> bool:
-        return False
 
     @property
     @abstractmethod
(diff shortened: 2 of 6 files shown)


neuralpp/symbolic/basic_expression.py
```diff
@@ -55,7 +55,7 @@ class BasicAtomicExpression(BasicExpression, AtomicExpression, ABC):
     def __init__(self, atom: Any, expression_type: Optional[ExpressionType] = None):
         if expression_type is None and not isinstance(atom, ExpressionType):
             # try to infer type for atom
-            if isinstance(atom, Callable):
+            if callable(atom):
```
hsiang-wu, line 58: `isinstance(..., Callable)` and `isinstance(...,type(Callable[...]))` are different (see later comments). This is an equivalent expression of the former, I made this change to avoid confusion with the latter.

neuralpp/test/quick_tests/symbolic/expression_test.py
```diff
@@ -49,7 +49,7 @@ def expression_factory(request):
 def test_constant(expression_factory):
     # Constant can be anything
     constant_one = expression_factory.new_constant(1)
-    constant_abc = expression_factory.new_constant("abc", Callable[[int], int])
+    constant_abc = expression_factory.new_variable("abc", Callable[[int], int])
```
hsiang-wu, line 52: fix bug. Uninterpreted functions should be variable instead of constant.

neuralpp/symbolic/expression.py
```diff
@@ -203,9 +197,8 @@ def _convert(cls, from_expression: Expression) -> Expression:
                     f"invalid from_expression {from_expression}: {type(from_expression)}"
                 )
 
-
     def get_return_type(self, number_of_arguments: int) -> ExpressionType:
-        if not isinstance(self.type, Callable):
+        if not isinstance(self.type, type(Callable[..., Any])):
```
hsiang-wu, line 201: E.g., `isinstance(int, Callable)` is `True` (because we can write `int()`), but `isinstance(int, type(Callable[..., Any]))` is `False`.

neuralpp/symbolic/z3_expression.py
```diff
@@ -104,7 +104,12 @@ def new_constant(
 
     @classmethod
     def new_variable(cls, name: str, type_: ExpressionType) -> Z3Variable:
-        z3_var = z3.Const(name, type_to_z3_sort(type_))
+        if isinstance(type_, type(Callable[..., Any])):
```
hsiang-wu, line 107: add supports for uninterpreted function as a z3 variable.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -81,6 +81,30 @@ def __init__(
         self._sympy_object = sympy_object
         self._type_dict = type_dict
 
+    def replace(self, from_expression: Expression, to_expression: Expression) -> Expression:
```
hsiang-wu, line 84: moved `replace()` from `SymPyFunctionApplicationInterface` to here because it should be in `SymPyExpression` if it's about sympy specific replace().

neuralpp/symbolic/sympy_expression.py
```diff
@@ -275,7 +296,7 @@ def new_quantifier_expression(
         # Unlike for FunctionApplication where the non-constructable is the exception,
         # here only a few SymPy-backed quantifier expression can be constructed from a general interface.
         # Operation is limited to sum and product, and constrain can only be a range.
-        raise NotImplementedError()
+        raise FutureWarning("Not Implemented")
```
hsiang-wu, line 299: Otherwise, PyCharm gives "new_quantifier_expression not implemented" warnings to all subclasses.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -515,13 +528,27 @@ def from_sympy_function_and_general_arguments(
         else:
             return SymPyFunctionApplication(sympy_object, type_dict)
 
-    def __new__(
-        cls, sympy_object: sympy.Basic, type_dict: Dict[sympy.Basic, ExpressionType]
+
+class SymPyPoly(SymPyExpression):
```
hsiang-wu, line 532: It is not a FunctionApplication.

neuralpp/symbolic/expression.py
```diff
@@ -117,22 +122,11 @@ def syntactic_eq(self, other) -> bool:
...
-                     FunctionApplication(function=Constant(value=functions.identity), arguments=identity_arguments),
-                     the_other
-                 ) | \
-                 (
-                     the_other,
-                     FunctionApplication(function=Constant(value=functions.identity), arguments=identity_arguments)
-                 ):
-                assert len(identity_arguments) == 1
-                return identity_arguments[0].syntactic_eq(the_other)
+            case (SymPyPoly(poly=poly), the_other) | (the_other, SymPyPoly(poly=poly)):
```
hsiang-wu, line 128: We still need a special case for Poly here, since `Poly` should be invisible. I believe `Poly(x+1)` should be defined as syntactically equal to `x+1`.

rodrigodesalvobraz: I think that is a good solution for now, but I think we will need to redesign `syntactic_eq` to be more modular at some point. In principle, `Expression` should not know about all its subclasses; people should be able to define their own new subclasses of `Expression` without needing to modify `expression.py`. So `syntactic_eq` should be an abstract method and each class takes care of its own comparisons. One trouble is that some classes may not know about newer classes they can be compared against (for example `FunctionApplication` can be compared to `SymPyPoly` but it does not know about the latter). One way to do that would be for each implementation of `A.syntactic_eq(B)` to run `B.syntactic_eq(A)` if `B` is an unknown class. We can extend the return type of the method to `None` if it does not know how to compare. So when confronted with an unknown class, the method calls the inverse order and, if it still gets `None`, then it finally returns `None`. I think I will take a stab at implementing this.

rodrigodesalvobraz: I think that is a good solution for now, but I think we will need to redesign `syntactic_eq` to be more modular at some point. In principle, `Expression` should not know about all its subclasses; people should be able to define their own new subclasses of `Expression` without needing to modify `expression.py`. So `syntactic_eq` should be an abstract method and each class takes care of its own comparisons. One trouble is that some classes may not know about newer classes they can be compared against (for example `FunctionApplication` can be compared to `SymPyPoly` but it does not know about the latter). One way to do that would be for each implementation of `A.syntactic_eq(B)` to run `B.syntactic_eq(A)` if `B` is an unknown class. We can extend the return type of the method to `None` if it does not know how to compare. So when confronted with an unknown class, the method calls the inverse order and, if it still gets `None`, then it finally returns `None`. I think I will take a stab at implementing this.

rodrigodesalvobraz (approved):
Thanks, nicely done redesign!

PR #51, opened 2022-09-15
branch pruning examples

hsiang-wu:
@rodrigodesalvobraz and I discussed a bit an example where z3 is helpful to prune branches. We decided to go with an example of (the integration of) the square of normal.

Here's a commit implementing this, with `evaluation_general_with_result` we can see the function prints the correct results for both [-20,20] (≈[-oo,oo]) and [-20,1] (≈[-oo,1]).

However, in this example, we don't even need z3: for simple pruning like `[-4,-13/4] & [13/4,4]`, sympy will automatically reduce this to `False`, so even with `Z3SolverExpressionDummy` as context, we can see the pruning of branch.

### DIFF
--- a/neuralpp/experiments/pimt_experiment.py
+++ b/neuralpp/experiments/pimt_experiment.py
@@ -24,6 +24,9 @@ joint = get_normal_piecewise_polynomial_approximation(x, mu1, 1.0, generator=mu1
     get_normal_piecewise_polynomial_approximation(mu1, mu2, 1.0, generator=mu1) \
     * \
     get_normal_piecewise_polynomial_approximation(mu2, BasicExpression.new_constant(0.0), 1.0, generator=mu1)
+normal_square = get_normal_piecewise_polynomial_approximation(x, BasicExpression.new_constant(0.0), 1.0, generator=x) \
+                * \
+                get_normal_piecewise_polynomial_approximation(x, BasicExpression.new_constant(0.0), 1.0, generator=x)
 
 E1 = x < 0
 E2 = x >= 0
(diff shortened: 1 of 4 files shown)


neuralpp/symbolic/general_normalizer.py
```diff
@@ -126,6 +126,7 @@ def _normalize(self, expression: Expression, context: Z3SolverExpression,
                 if context.contains(index):
                     raise ValueError(f"context {context} should not contain index {index}")
                 if context.is_known_to_imply(~constraint):
+                    # simple unsatisfiable interval is eliminated here, e.g., [-4,0] & [1,2]
```
hsiang-wu, line 129: Add print here and we can see prunable intervals are eliminated here.

neuralpp/symbolic/sympy_expression.py
```diff
@@ -196,6 +196,8 @@ def symbolic_integral(
                 a = big_f.replace(index.sympy_object, lower_bound.sympy_object)
             with profiler.profile_section("compute diff"):
                 diff = b - a
+                if all([gen.is_number for gen in diff.gens]):
+                    diff = diff.expr
```
hsiang-wu, line 200: If all generators are numbers, we want to substitute the polynomial with a numerical value

neuralpp/experiments/pimt_experiment.py
```diff
@@ -123,6 +128,13 @@ def two_normals_expectation():
 
 
 if __name__ == "__main__":
+    evaluation_general_with_result("branch pruning example",
+                                   normalization_generator(lambda: basic_integral(x, Z3SolverExpression.from_expression(x > -20.0) & (x < 20.0), normal_square)),
+                                   {})
+    evaluation_general_with_result("branch pruning example-2",
```
hsiang-wu, line 134: The function `evaluation_general_with_result` can be used when we want to verify the result. However, it won't work for some larger examples due to some limitations of SymPy's C code generation; also for large examples, the conversion time from our library representation to a SymPy one (https://github.com/rodrigodesalvobraz/neuralpp/pull/51/files#diff-f980e5935011a0e635e5749a155ed7feae9a9ea006fd4415f48372b77dd87e86R73) could take too long.

PR #52, opened 2022-09-23
Trying to Set Up a Linter

CactusWin:


CactusWin:
I'm not sure if we have github actions set up, I'll revisit this at another time

